Issue #64, opened 2022-09-23
Move common dependencies up to workspace Cargo.toml


cdata:
The latest Rust release includes enhanced capabilities for Cargo workspaces. In particular, common dependencies across every crate can now be specified in the workspace's Cargo.toml and inherited by the manifest for each crate in the workspace: https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table

We should probably uplift the most common dependencies (such a `anyhow`, `cid`, `async_trait` etc.). In addition to cutting down on bloat in our crate-specific manifests, it will help us to keep our dependency versions in sync across all crates in the workspace.

igbanam:
I'm currently looking into this. The ask is for "common dependencies". After reading through the workspaces doc linked, seems to me like all dependencies can be at the workspace level. How do we feel about that?

PR #65, opened 2022-09-27
Refactor crate dependencies

src255:
Move common dependencies to workspace dependencies. 

**Note:** some dependency versions of the form "X.Y" are now "~X.Y" to be consistent with other similar imports.


cdata:
Thank you for making this contribution @src255 ! This is relevant to #64 .

I admit, I hadn't read the documentation thoroughly. In the issue description, I speculated that this change would cut down on bloat. But, sadly that is not the case (since you must still inherit the dependency explicitly in the crate's manifest).

That said, this change still seems valuable for ensuring that our dependency versions are consistent across the workspace.

cdata:
@src255 looks like we have a build failure: https://github.com/subconsciousnetwork/noosphere/actions/runs/3132929628/jobs/5095833286

Please take a look, and I'll check back here when you have a resolution!

src255:
Apologies, I introduced a typo in *rust/noosphere/Cargo.toml* during refactoring. I think this was the only issue.

**rust/noosphere/Cargo.toml**
```diff
-ed25519.workspace = true
+ed25519-zebra.workspace = true
```

cdata:
Oh, one more thing: we require signed commits as a merge policy. Would you be willing to sign your commits?

https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

cdata (dismissed):
Thank you so much for taking on this task! :beers:

src255:
No problem! I just moved **axum** and **tower-http** from the workspace dependencies back to the individual Cargo.toml files.

PR #71, opened 2022-10-04
Introduce `noosphere-cli` crate

### LINKED ISSUE #63: Offer a CLI tool to enable easy, manual interaction with the Noosphere
Our ambition is for Noosphere to be consumed via a browser-like client such as Subconscious. But, Noosphere is also a generalized protocol for content distribution and addressing. To support basic flows that use Noosphere for arbitrary content, we should offer a bare-bones CLI tool that enables actions like sphere curation, gateway syncing and building edges across the sphere graph.

The CLI will probably be a Git-like experience, with live files rendered to a working tree. Saving updates to files to the sphere is a manual operation, as is syncing with a gateway.

Additionally, content from other spheres may eventually be rendered to the file system for the user as well, but for now we will leave that as out of scope.

cdata:
This change introduces the `noosphere-cli` crate, which contains our initial work to offer an accessible CLI client that is compatible with any Noosphere gateway, and may synchronize data between any other Noosphere clients and a folder on the local filesystem.

The `noospher-cli` builds a program called `orb` which exposes a variety of helpful, high-level operations to the user:

```
❯ orb
A CLI tool for saving, syncing and sharing content to the Noosphere

Usage: orb <COMMAND>

Commands:
  config   Read and manage configuration values for a local sphere
  key      Create and securely manage personal keys
  sphere   Create, join or share access to a sphere
  auth     Manage access to a sphere by holders of other keys
  status   Show details about files in the sphere directory that have changed since the last time the sphere was saved
  diff     If a difftool is configured, show a diff between files on disk and saved versions in the sphere
  save     Saves changed files to a sphere, creating and signing a new revision in the process; does nothing if there have been no changes to the files since the last revision
  sync     Synchronizes the local sphere with the copy in a configured gateway; note that this is a "conflict-free" sync that may cause local changes to be overwritten in cases where two or more clients have made changes to the same files
  publish  Tell a configured gateway to update the published version of the sphere in the Noosphere name system
  help     Print this message or the help of the given subcommand(s)

Options:
  -h, --help  Print help information (use `--help` for more detail)

```

In this initial change, the following subcommands have been implemented: `key` and `sphere`.

We will follow-up with a change that introduces `auth` support, and then `save` and `sync` subcommands will come next.

A lot of the legwork to get us to this place included building and landing https://github.com/ucan-wg/rs-ucan/pull/28, as well as refactoring storage primitives to better support the high-level use case of the CLI. This should put us on a better footing to implement the remaining subcommands in short order. The following issues have been added to track follow-on work: #72, #73,  #74, #75, #76 and #77.

Fixes #63


gordonbrander:
@cdata `orb ponder` should be an alias for `orb status` 🧙‍♂️

PR #84, opened 2022-10-06
(DRAFT) Initial WIP on `noosphere-name-service`

jsantell:
Initial WIP on `noosphere-name-service` a crate that handles both providing and querying a DHT to resolve sphere DID mapping to latest, verifiable CID.

Current status: record propagation/serving is not currently working, need to dive more into `libp2p::kad`'s implementation.

Brief description of how it works:
* On a "server", run `noosphere-name-system daemon --keyname $KEYNAME --pinned $PINNED_TOML_PATH`
  * Keys must already be configured (e.g. in `~/.noosphere/keys/$KEYNAME.{public|private}`)
  * `$PINNED_TOML_PATH` is a toml file with key/value pairs to serve (see `examples/pinned.toml`). This TOML store is only for testing really. These values in `pinned.toml` were generated from a local sphere DID, and a signed UCAN auth that should be authenticated, but no verification of actual values implemented yet.
  * Currently fails to propagate value; lots of handlers at the moment purely for logging.
 * In a "client", run `noosphere-name-system query get $SPHERE_DID`
   * `$SPHERE_DID` should be one of the keys being provided by the server, e.g. in `$PINNED_TOML_PATH`
 
Note libp2p requires protobufs.
* Linux apt: `sudo apt install protobuf-compiler cmake`
* MacOS Homebrew: `brew install protobuf cmake`


rust/noosphere-name-system/src/cli.rs
```diff
@@ -0,0 +1,152 @@
...
+use noosphere::authority::generate_ed25519_key;
+use noosphere::authority::restore_ed25519_key;
+use noosphere_cli::native::workspace::Workspace;
+use noosphere_name_system::NameSystem;
+use std::fs;
+use tokio;
+use tokio::signal;
+use toml;
+/// @TODO these materials should be exposed in noosphere::authority
+use ucan_key_support::ed25519::Ed25519KeyMaterial;
```
jsantell, line 13: @cdata WDYT about exposing key details in `noosphere::authority`? Though more UCAN integration will be needed here (in name-system, not specifically this CLI dev shell)

rust/noosphere-name-system/examples/pinned.toml
```diff
@@ -0,0 +1,3 @@
+[test_sphere]
```
jsantell, line 1: This TOML with pinned keys is purely for dev/testing. Maybe some existing infra in `noosphere-gateway`, a sqlite table, or simple files to track "pinned content" (using IPFS parlance) while running the node. Any insight here @cdata ?

The CLI was originally an example, hence this toml file being here. Probably should move it back to being an example. Depends on what the final interface looks like with the gateway. Sort of inspired by (mostly now deprecated) kubo CLI for interfacing with the DHT via low level commands for debugging/customization.

rust/noosphere-name-system/src/cli.rs
cdata: There are some generalized helpers that we use across the crates for generating and managing these keys:

https://github.com/subconsciousnetwork/noosphere/blob/88b08e3dc6e9df8386ed109b81bb38c8f06ceb4f/rust/noosphere/src/authority/key_material.rs#L16-L38

I would be happy for us to expand on these. Although, I would shy away from re-exporting implementation directly from other crates unless it seems like we really need to.

rust/noosphere-name-system/examples/pinned.toml
cdata: Yah, use [`SphereDb`](https://github.com/subconsciousnetwork/noosphere/blob/88b08e3dc6e9df8386ed109b81bb38c8f06ceb4f/rust/noosphere-storage/src/db.rs#L142-L159) or any [`Store`](https://github.com/subconsciousnetwork/noosphere/blob/main/rust/noosphere-storage/src/interface.rs#L40-L55)-implementing construct. They implement [`KeyValueStore`](https://github.com/subconsciousnetwork/noosphere/blob/88b08e3dc6e9df8386ed109b81bb38c8f06ceb4f/rust/noosphere-storage/src/interface.rs#L185-L213), so you can `set_key`, `get_key` for configuration like this.

In the big picture, this data is going to come from the `Sphere` data structure itself, which means it will ultimately come from one of these stores (although it will be in a block, not a k/v entry).

jsantell:
Handling in follow up PRs: #93

PR #89, opened 2022-10-08
SphereFS returns error when saving an empty change (fixes #85)

### LINKED ISSUE #85: `SphereFs` should give error result when saving an empty change
Currently, `SphereFs` will happily save a new revision of the sphere even if you make zero changes to it. We should give an error result instead with a topical message for the caller.

SebastianZimmer:
Make `SphereFS::save` return an error when there is neither a non-empty mutation nor any additional headers to save.

Issue: https://github.com/subconsciousnetwork/noosphere/issues/85

### DIFF
--- a/rust/noosphere-fs/src/fs.rs
+++ b/rust/noosphere-fs/src/fs.rs
@@ -276,13 +276,17 @@ where
         additional_headers: Option<Vec<(String, String)>>,
     ) -> Result<Cid> {
         let sphere = Sphere::at(&self.sphere_revision, &self.db);
-
-        // TODO(#85): Error result if mutation is empty and also additional headers are empty
-
-        let mut revision = sphere.try_apply_mutation(self.require_mutation()?).await?;
+        let mutation = self.require_mutation()?;
+        let mut revision = sphere.try_apply_mutation(mutation).await?;
 
         if let Some(mut headers) = additional_headers {
-            revision.memo.headers.append(&mut headers);
+            if headers.len() > 0 {
+                revision.memo.headers.append(&mut headers);
+            } else if mutation.is_empty() {
+                return Err(anyhow!("No changes to save"))
+            }
+        } else if mutation.is_empty() {
+            return Err(anyhow!("No changes to save"))
         }
 
         let new_sphere_revision = revision.try_sign(credential, proof).await?;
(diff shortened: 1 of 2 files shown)


rust/noosphere-fs/src/fs.rs
```diff
@@ -277,7 +277,12 @@ where
     ) -> Result<Cid> {
         let sphere = Sphere::at(&self.sphere_revision, &self.db);
 
-        // TODO(#85): Error result if mutation is empty and also additional headers are empty
+        let mutation_to_save = self.mutation.get().is_some() && !self.mutation.get().expect("").is_empty();
```
cdata, line 280: Since `mutation` is a `OnceCell`, there is a chance that it has not been initialized. There is a helper function on this struct that we use two lines below (L287) to get an error result in case the `mutation` isn't initialized.

We should move that up into a `let` binding so that we only do the check once:

```rs
let mutation = self.require_mutation()?;
```

Then you can avoid writing `self.mutation.get()` multiple times.

rust/noosphere-fs/src/fs.rs
```diff
@@ -277,7 +277,12 @@ where
...
+        let headers_to_save = additional_headers.is_some() && additional_headers.as_ref().expect("").len() > 0;
```
cdata, line 281: As a rule of thumb, we avoid using `expect` outside of tests (although there are perfectly valid non-test usages for it). On an `Err` result or `None` option, `expect(...)` will panic. But, panic is meant to indicate an unrecoverable error and is not appropriate if we think the caller may be able to recover. Instead, consider using a conditional to progressively unwrap the value and then throw the error later:

```rs
if let Some(mut headers) = additional_headers {
  revision.memo.headers.append(&mut headers);
} else if (mutation.is_empty()) {
  return Err(anyhow!("No changes to save"))
}
```

cdata (changes requested):
Thanks so much for taking on this issue @SebastianZimmer :raised_hands: I left a few pieces of minor feedback, and a test would be welcome as well

SebastianZimmer:
Thanks for the detailed feedback, @cdata. Such constructive and helpful comments are an ideal opportunity for me to learn and improve. 

I have updated the commit and added tests.

rust/noosphere-fs/src/fs.rs
SebastianZimmer: Thanks for the hints. I have updated this. Your suggestion does not check if the vector is there but empty. I have added a guard for this in my updated commit. If this is undesired behavior, I will remove this condition and the associated test again.

cdata: Yah, thanks for the callout there. It's a weird case where the vector is empty but the value is `Some(...)`, but good to handle it.

rust/noosphere-fs/src/fs.rs
```diff
@@ -276,13 +276,17 @@ where
...
-
-        let mut revision = sphere.try_apply_mutation(self.require_mutation()?).await?;
+        let mutation = self.require_mutation()?;
+        let mut revision = sphere.try_apply_mutation(mutation).await?;
 
         if let Some(mut headers) = additional_headers {
-            revision.memo.headers.append(&mut headers);
+            if headers.len() > 0 {
+                revision.memo.headers.append(&mut headers);
+            } else if mutation.is_empty() {
```
cdata, line 285: Maybe we could try a `match` to clean up the conditions a little? Something like...

```rs
match additional_headers {
  Some(mut headers) if headers.len() > 0 => {
    // Append
  },
  _ if mutation.is_empty() => {
    // Error
  }
}
```

cdata (approved):
I'm happy with this, but if you want to go the extra mile there may be a cleanup opportunity.

Thank you so much for the contribution! :beers:

rust/noosphere-fs/src/fs.rs
cdata: If interested, please feel free to clean this up in a small PR. Otherwise I'm sure I can get to it eventually. Thanks again!

PR #92, opened 2022-10-14
Implement `orb config` and `orb serve` subcommands

### LINKED ISSUE #24: Implement gateway fetch handler
In #22 we completed the push half of the Git-like fetch/push synchronization scheme. Now we need to implement the fetch half, in both client and gateway. Given a sphere identity and a base CID in its lineage, the gateway should respond with a bundle containing the blocks the client needs to manifest the remaining lineage.

### LINKED ISSUE #74: Implement `orb config` subcommand
The `config` subcommand enables a user to set workspace-level key/value configurations that control how the `orb` command behaves when used in that workspace. For example: `orb config set gateway $URL` can be used to configure the gateway to use for synchronization.

### LINKED ISSUE #91: Implement `orb serve` subcommand
After some discussion, we've decided that the `noosphere-gateway` crate should be deprecated and the associated functionality be moved into an `orb serve` subcommand of the `noosphere-cli`.

cdata:
This change proposes basic implementations for the `orb config` and `orb serve` subcommands.

The substance of the `noosphere-gateway` crate has been relocated under the `noosphere-cli` crate and offered as an `orb serve` subcommand. This command launches a gateway that can be accessed by clients (including `orb`) that both receives and serves blocks related to a sphere. The `noosphere-gateway` crate has been removed from this project. Additionally, the fetch route of the gateway is implemented as of this change.

We also implemented `orb config`, which is necessary in order for a user to be able to establish a DID-based recognition between the gateway and any clients. The `orb config` command enables a user to configure the URL of a gateway (if configuring a client), and also the "counterpart" sphere DID for both gateways and clients.

Fixes #24 
Fixes #74 
Fixes #91


jsantell (approved):
Looks good, great timing too as I'm starting to use Workspace :+1:

PR #93, opened 2022-10-17
Initial work on `noosphere-p2p`, the libp2p-based DHT underpinings of a name system.

jsantell:


cdata:
Very exciting! I'll take a look at this starting today, although it might take me until tomorrow to get through it.

rust/noosphere-p2p/src/dht/config.rs
```diff
@@ -0,0 +1,178 @@
...
+    pub listening_address: libp2p::Multiaddr,
+    /// How frequently, in seconds, the DHT attempts to
+    /// dial peers found in its kbucket. Outside of tests,
+    /// should not be lower than 5 seconds.
+    pub peer_dialing_interval: u64,
+    pub query_timeout: u32,
+}
+
+impl DHTConfig {
+    pub fn bootstrap_interval(mut self, interval: u64) -> Self {
```
jsantell, line 32: These setters should be `&mut self` for easier non-one-liner configs

rust/noosphere-p2p/src/dht/node.rs
```diff
@@ -0,0 +1,180 @@
...
+            config,
+            state: DHTStatus::Active,
+            client: Some(client),
+            thread_handle: Some(thread_handle),
+        })
+    }
+
+    /// Spawns a networking thread with a DHT node, returning the
+    /// corresponding [DHTNode].
+    pub fn spawn(config: DHTConfig) -> Result<DHTNode, DHTError> {
```
jsantell, line 45: This could probably just be removed in lieu of `new`, but did not want to trivialize starting up a listener for a network

rust/noosphere-p2p/src/dht/processor.rs
```diff
@@ -0,0 +1,582 @@
...
+            }
+            Err(e) => {
+                $message.respond(Err(e));
+            }
+        };
+    };
+}
+
+/*
+macro_rules! take_requests {
```
jsantell, line 73: Will remove

rust/noosphere-p2p/src/dht/processor.rs
```diff
@@ -0,0 +1,582 @@
...
+        f.debug_struct("DHTNode")
+            .field("peer_id", &self.peer_id)
+            .field("config", &self.config)
+            .finish()
+    }
+}
+
+macro_rules! dht_trace {
+    ($self:expr, $event:expr) => {
+        trace!("{:#?}::{:#?}", $self.peer_id.to_base58(), $event);
```
jsantell, line 51: Tracing in this event handler here is kind of a mess currently

rust/noosphere-p2p/src/dht/processor.rs
```diff
@@ -0,0 +1,582 @@
...
+        self.swarm
+            .behaviour_mut()
+            .kad
+            .bootstrap()
+            .and_then(|_| Ok(()))
+            .map_err(|_| DHTError::NoKnownPeers)
+    }
+}
+
+impl Drop for DHTProcessor {
```
jsantell, line 578: Will remove

rust/noosphere-p2p/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+[dependencies]
+anyhow = "^1"
+tracing = "0.1"
+
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+futures = "0.3.1"
+tokio = { version = "1.15", features = ["sync", "io-util", "io-std", "macros", "rt", "rt-multi-thread"] }
+ucan = { version = "0.7.0-alpha.1" }
+ucan-key-support = { version = "0.7.0-alpha.1" }
+libp2p = { version = "0.48.0", default-features = true, features = [ "identify", "dns-tokio", "tcp-tokio", "websocket", "noise", "mplex", "yamux", "kad" ] }
```
jsantell, line 16: TODO: Try with `default-features = false`

rust/noosphere-p2p/src/dht/config.rs
jsantell: I take that back, as the config object is consumed by `DHTNode`.

jsantell:
Revisions:
* Cleaned up the cargo manifests:
  * Remove the deps for the previous iteration of "name-system" (CLI, noosphere-cli stuff)
  * Updating libp2p from 0.48.0 to 0.49.0! There are a handful of feature flag changes and deprecated network behaviours in the latest release, so this cleans up some TODOs and flags.
  * Turn off `default-features` for libp2p.
* Update tests:
  * Skip integration tests in wasm builds.
  * Use real IP4 loopback addresses (`/ip4/127.0.0.1/tcp/42533`) instead of memory addresses (`/memory/123412341234`) for testing.
* API stuff:
  * Changed `DHTNode::spawn` to `DHTNode::new` -- it's attractive to use an intentionally different name to indicate some sort of non-trivial instantiation is occurring (e.g. a networking thread), but preferring the idioms and good docs for now.
  * `DHTConfig`: due to this struct containing immutable state that's used in multiple threads, in lieu of a builder pattern, I have setter methods on `DHTConfig` in a [consuming-builder style](https://rust-lang.github.io/api-guidelines/type-safety.html#consuming-builders). Anything I should change here for consistency throughout Noosphere?
  * Changed some `Option<T>`s to `T` in `DHTNode`; as the network starts upon instantiation (`DHTNode::new`) and is useless after termination (e.g., restarting means creating a new instance).
* Removed some garbage

rust/noosphere-p2p/src/dht/transport.rs
```diff
@@ -0,0 +1,35 @@
...
+/// Creates the Transport mechanism that describes how peers communicate.
+/// Currently, mostly an inlined form of `libp2p::tokio_development_transport`.
+pub(crate) fn build_transport(
+    keypair: &libp2p::identity::Keypair,
+) -> Result<Boxed<(PeerId, StreamMuxerBox)>, io::Error> {
+    let transport = {
+        let dns_tcp = dns::TokioDnsConfig::system(tcp::TokioTcpTransport::new(
+            tcp::GenTcpConfig::new().nodelay(true),
+        ))?;
+        let ws_dns_tcp = websocket::WsConfig::new(dns::TokioDnsConfig::system(
```
cdata, line 16: Just curious: does the WebSocket transport help for cases beyond peering with browser nodes?

rust/noosphere-p2p/src/dht/types.rs
```diff
@@ -0,0 +1,130 @@
...
+            num_pending: c.num_pending(),
+            num_established: c.num_established(),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub enum DHTRequest {
+    Bootstrap,
+    //WaitForPeers(usize),
```
cdata, line 36: Is this for the case where you want to ensure a peer quorum of some size?

rust/noosphere-p2p/tests/integration_test.rs
```diff
@@ -0,0 +1,103 @@
...
+/// and ensuring clients become peers.
+#[test_log::test(tokio::test)]
+async fn test_dhtnode_bootstrap() -> Result<(), DHTError> {
+    let num_clients = 5;
+    let (mut bootstrap_nodes, mut client_nodes) = initialize_network(1, num_clients).await?;
+    let bootstrap = bootstrap_nodes.pop().unwrap();
+
+    for info in swarm_command(&mut client_nodes, |c| c.network_info()).await? {
+        assert_eq!(info.num_peers, num_clients);
+        // @TODO the number of connections seem inconsistent??
```
cdata, line 49: Same feedback re: `TODO` comments above

cdata (changes requested):
This is awesome. I was able to check it out and build it locally, while is also awesome :sweat_smile: 

My feedback is fairly minor. I think this is a great start and we can move forward from this foundation.

rust/noosphere-p2p/src/dht/behaviour.rs
```diff
@@ -0,0 +1,75 @@
...
+
+impl From<IdentifyEvent> for DHTEvent {
+    fn from(event: IdentifyEvent) -> Self {
+        DHTEvent::Identify(event)
+    }
+}
+
+pub type DHTSwarmEvent = SwarmEvent<
+            <DHTBehaviour as swarm::NetworkBehaviour>::OutEvent,
+            <<<DHTBehaviour as swarm::NetworkBehaviour>::ConnectionHandler as IntoConnectionHandler>::Handler as ConnectionHandler>::Error>;
```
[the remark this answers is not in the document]
jsantell: Yeah :cry: this SwarmEvent type expanded once introducing a second "sub behaviour" (Identify)

rust/noosphere-p2p/src/dht/transport.rs
jsantell: Probably can be removed at least for now, a peer can offer multiple addresses at once, and unclear to me yet how the web version works in context

rust/noosphere-p2p/src/dht/types.rs
jsantell: In lieu of "events" (messages from processor to client), this was a work around to get tests just running with some form of confidence. I think "events" would remove the need for this (with maybe the client side still having a helper function for it, maybe just in tests, built off of said Events). Other DHT commands like putting records can have a quorum constraint as well

rust/noosphere-p2p/src/dht/config.rs
```diff
@@ -0,0 +1,107 @@
...
+pub struct DHTConfig {
+    /// If bootstrap peers are provided, how often,
+    /// in seconds, should the bootstrap process execute
+    /// to keep routing tables fresh.
+    pub bootstrap_interval: u64,
+    /// Peer addresses to query to update routing tables
+    /// during bootstrap. A standalone bootstrap node would
+    /// have this field empty.
+    pub bootstrap_peers: Vec<libp2p::Multiaddr>,
+    pub keypair: libp2p::identity::Keypair,
```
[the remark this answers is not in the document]
cdata: As mentioned elsewhere, the rule of thumb for me is that if we are defining the interface, we should err on the side of constructs that we already use elsewhere in our own interfaces. It's not always an option, but at least in the case of keys we have a thing we use everywhere.

rust/noosphere-p2p/src/dht/processor.rs
```diff
@@ -0,0 +1,483 @@
...
+            KademliaEvent::InboundRequest { request } => match request {
+                kad::InboundRequest::FindNode {
+                    num_closer_peers: _,
+                } => {}
+                kad::InboundRequest::GetProvider {
+                    num_closer_peers: _,
+                    num_provider_peers: _,
+                } => {}
+                kad::InboundRequest::AddProvider { record: _ } => {}
+                kad::InboundRequest::GetRecord {
```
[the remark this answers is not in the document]
jsantell: Actually, the VPS/e2e tests show GetRecord can fetch the data from peers without a local cache, but doesn't illustrate a deeper network "search"

rust/noosphere-p2p/src/dht/behaviour.rs
```diff
@@ -0,0 +1,76 @@
...
+
+#[derive(NetworkBehaviour)]
```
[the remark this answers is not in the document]
jsantell: Adding this entry fixed it!
```
"libp2p_swarm_derive": [
      "NetworkBehaviour"
],
```

rust/noosphere-p2p/src/dht/behaviour.rs
```diff
@@ -0,0 +1,76 @@
+use crate::dht::DHTConfig;
+use libp2p::{
+    // In the next version of libp2p-identify 0.40.0 (the subpackage), these values
```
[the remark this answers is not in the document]
jsantell: Out of date, now that we're using libp2p 0.49.0 here

rust/noosphere-p2p/src/dht/behaviour.rs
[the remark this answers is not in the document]
jsantell: Nevermind, it did not :cry:

jsantell:
Updated based off of feedback; noteworthy changes:

* Added `DHTNodeBuilder` to offer a more noosphere-esque interface rather than libp2p (e.g. `Ed25519KeyMaterial` and `String` instead of `libp2p::Keypair` and `libp2p::Multiaddr` and provide an option that doesn't eagerly generate a key.
  * `DHTConfig::default` remains mostly for testing. The non-builder creation methods (`DHTNode::new`) are still public, specifically for integration tests. Could move `DHTConfig::default` exclusively to tests.
  *  The `DHTNodeBuilder` example docs and integration tests illustrate the challenges of using non-libp2p objects as an interface, when getters/computed values are needed to populate other nodes (e.g. connecting multiple bootstrap nodes to each other, requiring addresses for each to be shared with each node). Mostly an issue with `Multiaddr` properties (`listening_address`, `bootstrap_peers`). The public interface of `DHTNode` could convert these internal `Multiaddr`s back to String, but that would ultimately get converted back into a `Multiaddr` at some point. As an alternative to `String` interfaces, essentially a typeless `Multiaddr`, we could introduce a different type, like `TcpSocket` and `Ipv4Addr`, but that same translation will still need to be done when wiring multiple nodes together.
* Remove web socket transport

cdata (approved):
This is a very exciting start to something extremely cool :beers: awesome work @jsantell !

PR #97, opened 2022-10-19
Centralize common dependencies as workspace dependencies

igbanam:


igbanam:
**This is a WIP PR.**

I think we should use this PR to discuss how we want this to fulfil #64.

What's the convention in the Rust community. Should only the common packages live as workspace dependencies? I see that it's a possibility to have all dependencies as workspace dependencies and selectively bring them into packages.

igbanam:
I'll close this and re-open. Forgot to create a branch locally

PR #98, opened 2022-10-20
Refactor common dependencies into workspace dependencies

### LINKED ISSUE #64: Move common dependencies up to workspace Cargo.toml

igbanam:
Closes #64


igbanam:
**This is a WIP PR.**

I think we should use this PR to discuss how we want this to fulfil #64.

What's the convention in the Rust community. Should only the common packages live as workspace dependencies? I see that it's a possibility to have all dependencies as workspace dependencies and selectively bring them into packages.

cdata:
Thank you so much for taking this on. This change has a potentially large scope. I would be happy for you to attempt it over a series of PRs if that makes it easier - I'll leave it to your discretion.

cdata:
> What's the convention in the Rust community. Should only the common packages live as workspace dependencies?

This feature is _very_ new to Rust, so I don't know that there are any common conventions yet. For our purposes, there are packages where it is obvious that we should be doing this: `anyhow`, `cid`, `tracing`, `rand`, `libipld-core` etc. and then there is a long tail of packages where it is less obvious.

I think it would be good to start with at least the most commonly re-used packages. After we set those up, we can incrementally move up to the workspace level on a case-by-case basis as time and opportunity allows for it.

jsantell:
There have been some crate-level changes over the last few weeks, having this supported would be great! Is there any info or help I could provide?

Now that we have some release infrastructure for individual crates, could this be used for our own crates as well? (e.g. `noosphere` in `noosphere-ns`)

igbanam:
I have a weird problem with this one. `main` — as of today — does not build tests for me. I think I traced the problem down to [these lines](https://github.com/subconsciousnetwork/noosphere/blob/fad2aa2c9c48d8eed723947b499940b4a17b0536/rust/noosphere/src/platform.rs#L60-L66). I'm on a Mac, this means the code path here is not available on my machine. But those are needed [here](https://github.com/igbanam/noosphere/blob/fad2aa2c9c48d8eed723947b499940b4a17b0536/rust/noosphere/src/sphere/builder.rs#L334-L336) 

How do I resolve this?

jsantell:
> I have a weird problem with this one. `main` — as of today — does not build tests for me. I think I traced the problem down to [these lines](https://github.com/subconsciousnetwork/noosphere/blob/fad2aa2c9c48d8eed723947b499940b4a17b0536/rust/noosphere/src/platform.rs#L60-L66). I'm on a Mac, this means the code path here is not available on my machine. But those are needed [here](https://github.com/igbanam/noosphere/blob/fad2aa2c9c48d8eed723947b499940b4a17b0536/rust/noosphere/src/sphere/builder.rs#L334-L336)
> 
> How do I resolve this?

Sorry for the slow response @igbanam ! We did recently get macOS and Windows builds working and in our CI, I wonder if the latest changes (which is admittedly a lot) will fix the issue you're seeing

PR #111, opened 2022-10-25
dev: Changes to DHTNode for a more Noosphere-specific interface

jsantell:
Changes to DHTNode for a more Noosphere-specific interface rather than a heavily libp2p one. Consolidate some underlying libp2p components (swarm, behaviour, transport). Introduce an explicit DHTNode::run method to control when the processor starts, in order to configure bootstrap peers before the bootstrap process. Decouple the non-primitive/libp2p-specific configurations from DHTConfig.


cdata (approved):
Looks like a reasonable improvement over the first pass!

PR #116, opened 2022-10-27
Rename noosphere-p2p to noosphere-ns in preparation for crate publishing

jsantell:


cdata (approved):
Looks great :raised_hands: thanks for the quick turn-around!

PR #117, opened 2022-10-29
Add release automation

cdata:
This change introduces a Github Actions-based workflow for release automation that is built on [`release-please`](https://github.com/googleapis/release-please).

This workflow is stateful. When there are unreleased changes on main, it generates a candidate PR for the next release based on all the unreleased changes and their implications for the versions of our packages. If such a PR already exists, it will detect it and update in place. These PRs are intended to stick around and update until a release is ready to be made. Once merged, this step will detect the merge and generate Git tags and corresponding Github releases for each package being released. Only at that time will the rest of the steps in this workflow run.

After a release is made, the workflow builds artifacts and adds them to the Github releases they correspond to. Then, it publishes our crates to crates.io.

The release workflow depends on us using [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) so that it can decide how to update versions over time. Its version updating logic is based on the configuration in our repository, so for the time being it will bump minor versions for breaking changes. When we are ready to go for a 1.0, we can tweak this behavior on a per-package or global basis.

Note that the workflow is currently using a fork of `release-please`. In the course of implementing this workflow, I discovered a bug in how it bumps versions in Cargo manifests under some conditions that apply to us. I put up a PR with a fix for consideration upstream (https://github.com/googleapis/release-please/pull/1730). When it lands, we can switch off the fork.

In addition to the release workflow, I've taken a stab at optimizing our test workflow a little bit with (hopefully) better caching and pre-built binary installation of some external Rust tools we were previously compiling from source.


jsantell (approved):
Great work! Getting these releases working does a lot of heavy lifting, and a challenge to set up.

> The release workflow depends on us using [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) so that it can decide how to update versions over time. Its version updating logic is based on the configuration in our repository, so for the time being it will bump minor versions for breaking changes. When we are ready to go for a 1.0, we can tweak this behavior on a per-package or global basis.

* A guide on commits (`.github/CONTRIBUTIONS.md` ?) could be helpful in the future (explicitly what flags we care about e.g. `fix` `feat` `chore`).
* At least with the domains I'm familiar with, and still in flux, any issues pre-1.0 with a lot of breaking changes? Worst case scenario maybe is `noosphere-ns` version `0.100.0` though (or possibly minor-change bumping all crates?), though ultimately a consequence of experimental published crates.
* In favor of minimizing infra resources, is there anyway to only run the packaging on-demand e.g. "we want to release updates this month"? Improved caching certainly helps here (and with CI tests too). What publishing cadence do you have in mind in the current state, and in 6 months?

This is excellent, looking forward to getting this landed, and we can follow up with infra/contribution/release docs

cdata:
Thanks @jsantell - see my thoughts below.

> A guide on commits (.github/CONTRIBUTIONS.md ?) could be helpful in the future (explicitly what flags we care about e.g. fix feat chore).

Filed #118 :+1: 

> At least with the domains I'm familiar with, and still in flux, any issues pre-1.0 with a lot of breaking changes? Worst case scenario maybe is noosphere-ns version 0.100.0 though (or possibly minor-change bumping all crates?), though ultimately a consequence of experimental published crates.

For now, breaking changes will bump minor versions. I'm happy for us to be eager with breaking changes at this time, and I don't think it's an issue if we get to `0.112093.z` of any given crate ahead of `1.0.0`.

> In favor of minimizing infra resources, is there anyway to only run the packaging on-demand e.g. "we want to release updates this month"? Improved caching certainly helps here (and with CI tests too). What publishing cadence do you have in mind in the current state, and in 6 months?

If we have a use-case of on-demand builds, let's figure out what it is and layer that on as an enhancement to this change!

PR #122, opened 2022-11-01
dev: initial work on NameSystem, wrapping the underlying DHT network.

jsantell:
Implementation of the NameSystem propagating records through the DHT network.

* Top-level exports like `NameSystem`, `NameSystemBuilder` and `NSRecord` all have rather robust tests and docs. 
  * `NSRecord` is the struct being transmitted over the DHT. Rolling our own de/serializer due to cid's built-in serde implementation always writing to bytes. The structure is transmitted as JSON UTF-8. Flexible about details here, but is an area that will need to be "standardized" as part of the protocol. I imagine additional fields on here once we add UCAN.
  * `NameSystem` and `NameSystemBuilder` APIs are flexible, but attempt to mimic DNS.
  * Re-exports aren't ideal, but the multiformats `Cid` and `Multiaddr` are rather intrinsic to Noosphere/NS.
* [ns_test](https://github.com/subconsciousnetwork/noosphere/blob/namesystem/rust/noosphere-ns/tests/ns_test.rs) is a good example of functionality.
* Not yet implemented:
  * CLI demo; initial work stashed elsewhere for now, can integrate it fully into orb when ready
  * SphereIpld integration; initial work stashed elsewhere containing a `names` entry, can start this either with UCAN or CLI integration.
  * Any UCAN integration (hydration, validation)
  * Automatically propagating provided records at TTL interval


rust/noosphere-ns/src/name_system.rs
```diff
@@ -0,0 +1,195 @@
...
+        self.dht_set_record(identity, &record).await?;
+        self.hosted_records.insert(identity.to_owned(), record);
+        Ok(())
+    }
+
+    /// Gets the content Cid for the provided sphere identity.
+    pub async fn get_record(&mut self, identity: &String) -> Option<&Cid> {
+        // Round about way of checking for local valid record before
+        // querying the DHT network due to the borrow checker.
+        // https://stackoverflow.com/questions/70779967/rust-borrow-checker-and-early-returns#
```
jsantell, line 95: Not ideal, but an early return always holds on to ownership, even when the conditional is not triggered. This work-around works, but ideally we don't have to get the record twice from the HashMap. Any ideas @cdata ?

rust/noosphere-ns/src/name_system.rs
```diff
@@ -0,0 +1,195 @@
...
+        }
+
+        let record = NSRecord::new_from_ttl(link.to_owned(), self.ttl)?;
+        self.dht_set_record(identity, &record).await?;
+        self.hosted_records.insert(identity.to_owned(), record);
+        Ok(())
+    }
+
+    /// Gets the content Cid for the provided sphere identity.
+    pub async fn get_record(&mut self, identity: &String) -> Option<&Cid> {
```
jsantell, line 92: After more thought, this should probably return a `NSRecord`, especially as we'll be storing 2-3 fields. WDYT?

rust/noosphere-ns/src/name_system.rs
```diff
@@ -0,0 +1,195 @@
...
+            .collect();
+        try_join_all(pending_tasks).await?;
+        Ok(())
+    }
+
+    /// Propagates the corresponding managed sphere's content Cid on nearby peers
+    /// in the DHT network.
+    ///
+    /// Can fail if NameSystem is not connected or if no peers can be found.
+    pub async fn set_record(&mut self, identity: &String, link: &Cid) -> Result<()> {
```
jsantell, line 80: Similarly to `get_record`, this should also probably take a `NSRecord` instead of a `Cid` as we'll be propagating 2-3 values at least per record.

rust/noosphere-ns/src/name_system.rs
```diff
@@ -0,0 +1,195 @@
...
+use cid::Cid;
+use futures::future::try_join_all;
+use libp2p::Multiaddr;
+use std::collections::HashMap;
+use ucan_key_support::ed25519::Ed25519KeyMaterial;
+
+/// The [NameSystem] is responsible for both propagating and resolving address records
+/// in the Noosphere DHT. Hosted records can be set via `set_record(identity, link)`, propagating the
+/// record immediately, and repropagated every `propagation_interval` seconds. Records
+/// can be resolved via `get_record(identity)`.
```
jsantell, line 15: I have follow up doc changes here ^

rust/noosphere-ns/src/dht/swarm.rs
```diff
@@ -52,6 +52,7 @@ impl DHTBehaviour {
         let kad = {
             let mut cfg = KademliaConfig::default();
             cfg.set_query_timeout(Duration::from_secs(config.query_timeout.into()));
+            cfg.set_record_filtering(KademliaStoreInserts::FilterBoth);
```
jsantell, line 55: Rather than automatically storing propagated records from other nodes, require it to be manually added -- this allows us a hook to perform validation and lookups (not yet implemented, everything is added anyway lol)

rust/noosphere-ns/src/dht/processor.rs
```diff
@@ -199,6 +197,7 @@ impl DHTProcessor {
                     publisher: None,
                     expires: None,
                 };
+                info!("PUTTING RECORD {:#?}", record);
```
jsantell, line 200: Debug value, will remove

rust/noosphere-ns/src/lib.rs
```diff
@@ -1,5 +1,24 @@
+#[cfg(not(target_arch = "wasm32"))]
 #[macro_use]
 extern crate tracing;
 
 #[cfg(not(target_arch = "wasm32"))]
 pub mod dht;
+
+#[cfg(not(target_arch = "wasm32"))]
```
jsantell, line 8: Is there a way to group these `cfg` conditionals into a mod or something? Hella noisy

rust/noosphere-ns/src/name_system.rs
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    /// Clears out the internal cache of resolved records
+    /// for the matched identity. Returned value indicates whether
+    /// a record was successfully removed.
+    pub fn flush_records_for_identity(&mut self, identity: &String) -> bool {
+        self.resolved_records.remove(identity).is_some()
+    }
+
+    pub fn p2p_address(&self) -> Option<&Multiaddr> {
```
jsantell, line 132: Unused currently; previously used to log listening address in CLI, which seems useful, so leaving it here for now

rust/noosphere-ns/src/records.rs
```diff
@@ -0,0 +1,318 @@
...
+            .ok_or_else(|| anyhow!("Duration overflow."))?
+            .duration_since(UNIX_EPOCH)
+            .map_err(|e| anyhow!(e.to_string()))?
+            .as_secs();
+        Ok(Self { cid, expires })
+    }
+
+    /// Creates a new [NSRecord] from serialized bytes. See [NSRecord]
+    /// for serialization details.
+    pub fn from_bytes(bytes: Vec<u8>) -> Result<Self, Error> {
```
jsantell, line 75: I attempted to implement `TryFrom<Vec<u8>>` and `TryInto<Vec<u8>>` but the trait object types started getting out of control. Any thoughts on these bespoke methods versus well-known traits? Either way, these methods are nice to hide the serialization "codec" that needs to be shared amongst all DHT nodes.

rust/noosphere-ns/src/records.rs
```diff
@@ -0,0 +1,318 @@
...
+        match SystemTime::now().duration_since(UNIX_EPOCH) {
+            Ok(duration) => duration.as_secs() >= self.expires,
+            Err(_) => false,
+        }
+    }
+}
+
+impl fmt::Display for NSRecord {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        match serde_json::to_string(self) {
```
jsantell, line 99: Any issues with implementing this display as JSON? No expectations on how display strings are rendered, but the JSON form fits well to reduce the amount of potential formats this data can be displayed.

rust/noosphere-ns/src/records.rs
```diff
@@ -0,0 +1,318 @@
...
+            Ok(record_str) => write!(f, "{}", record_str),
+            Err(_) => write!(f, "{{ INVALID_RECORD }}"),
+        }
+    }
+}
+
+/// Serialization for NSRecords. While [cid::Cid] has built-in serde
```
jsantell, line 106: nit: if `Cid` is a top level re-export, docs should change `s/cid::Cid/Cid/g`

rust/noosphere-ns/src/records.rs
```diff
@@ -0,0 +1,318 @@
...
+        state.serialize_field("cid", &self.cid.to_string())?;
+        state.serialize_field("exp", &self.expires)?;
+        state.end()
+    }
+}
+
+/// Deserialization for NSRecords. While [cid::Cid] has built-in serde
+/// support under a feature flag, we roll our own to store the Cid
+/// as a string rather than bytes.
+/// For more details on custom deserialization: https://serde.rs/deserialize-struct.html
```
jsantell, line 124: nit: wrap link in `< >`

rust/noosphere-ns/tests/ns_test.rs
```diff
@@ -0,0 +1,110 @@
...
+    // Fetch record from network if local record is expired.
+    ns_2.set_record(&sphere_2_id, &sphere_2_cid_2).await?;
+    // Wait to ensure record is expired ;_;
+    wait_ms(1000).await;
+    let cid = ns_1.get_record(&sphere_2_id).await.expect("to be some");
+    assert_eq!(
+        cid, &sphere_2_cid_2,
+        "record found from network from local expired record"
+    );
+
```
jsantell, line 108: Any other scenarios to test here with the current implementation?

rust/noosphere-ns/src/dht/swarm.rs
jsantell: DHT Store also has its own expiration, which is not yet tied into the NS propagation timer

rust/noosphere-ns/tests/ns_test.rs
```diff
@@ -0,0 +1,110 @@
...
+use cid::{
+    multihash::{Code, MultihashDigest},
+    Cid,
+};
+use noosphere_core::authority::generate_ed25519_key;
+use noosphere_ns::{NameSystem, NameSystemBuilder};
+
+use utils::{create_bootstrap_nodes, wait_ms};
+
+fn new_cid(bytes: &[u8]) -> Cid {
```
cdata, line 14: We have a slightly better way to do this now: https://github.com/subconsciousnetwork/noosphere/blob/1a0625ad79330d35ca137361297318bdbf29137e/rust/noosphere-storage/src/encoding.rs#L23-L32

```rs
derive_cid::<RawCodec>(&bytes);
```

Also, conventionally we are using `Blake2b256` everywhere. Is `Sha2_256` a hard requirement?

cdata (commented):
This is a really crisp iteration over what has already been landed, great work! I have some general questions and a few small actionable items, but otherwise I'm happy for us to ship it.

rust/noosphere-ns/src/name_system.rs
jsantell: From yesterday's discussion, using a Ucan as the record will work well (or possibly if we need additional data from the NS, like a ttl, a wrapper around a Ucan string)

rust/noosphere-ns/src/lib.rs
jsantell: Beautiful

rust/noosphere-ns/src/builder.rs
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    /// How long, in seconds, until a network query times out.
+    pub fn query_timeout(mut self, timeout: u32) -> Self {
+        self.dht_config.query_timeout = timeout;
+        self
+    }
+
+    /// Default Time To Live (TTL) for records propagated to the network.
+    pub fn ttl(mut self, ttl: u64) -> Self {
```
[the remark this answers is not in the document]
jsantell: Yup, can mostly be replaced by Ucan's time-bounds, unless we want the NS to staple on its own TTL to the immutable ucan token (e.g. every 2 weeks, the NS pushes out a new record with new ttl) -- currently there's an implication as well on propagation interval (not yet implemented) but separate from the record representation

rust/noosphere-ns/src/dht/processor.rs
```diff
@@ -350,7 +348,13 @@ impl DHTProcessor {
...
                 kad::InboundRequest::PutRecord { source, record, .. } => match record {
                     Some(rec) => {
-                        if self.swarm.behaviour_mut().kad.store_mut().put(rec.clone()).is_err()
+                        if self
+                            .swarm
+                            .behaviour_mut()
+                            .kad
+                            .store_mut()
+                            .put(rec.clone())
+                            .is_err()
```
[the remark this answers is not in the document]
jsantell: I'll add a log, these are errors failing to save some data from another node which can only occur for invalid data, a full store, or a data that's too that generally feels more like pruning rather than an error-error

rust/noosphere-ns/src/name_system.rs
```diff
@@ -0,0 +1,202 @@
...
+            None
+        }
+    }
+
+    /// Queries the DHT for a record for the given sphere identity.
+    /// If no record is found, no error is returned.
+    ///
+    /// Returns an error if not connected to the DHT network.
+    async fn dht_get_record(&self, identity: &String) -> Result<(String, Option<NSRecord>)> {
+        let dht = self.dht.as_ref().ok_or_else(|| anyhow!("not connected"))?;
```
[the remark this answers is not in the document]
jsantell: I like it!

rust/noosphere-ns/src/records.rs
```diff
@@ -0,0 +1,320 @@
...
+            .map_err(|e| anyhow!(e.to_string()))?
+            .as_secs();
+        Ok(Self { cid, expires })
+    }
+
+    /// Creates a new [NSRecord] from serialized bytes. See [NSRecord]
+    /// for serialization details.
+    pub fn from_bytes(bytes: Vec<u8>) -> Result<Self, Error> {
+        let string = String::from_utf8(bytes).map_err(|e| anyhow!(e.to_string()))?;
+        serde_json::from_str(&string).map_err(|e| anyhow!(e.to_string()))
```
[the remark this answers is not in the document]
jsantell: No opinion I think on the JSON representation, just the quick and easy solution for now. If we do not need to "staple" NS data onto the record, this could simply be utf8 encoded Ucan

rust/noosphere-ns/src/records.rs
```diff
@@ -0,0 +1,320 @@
...
+};
+
+/// An [NSRecord] is a struct representing a record stored in the
+/// Noosphere Name System's DHT containing a [Cid] of the
+/// result, as well as a TTL expiration as seconds from Unix epoch.
+///
+/// # Serialization
+///
+/// When transmitting records across the network, they're encoded
+/// as JSON-formatted UTF-8 strings. The [NSRecord::to_bytes]
```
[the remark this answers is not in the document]
jsantell: From other comment: No opinion I think on the JSON representation, just the quick and easy solution for now. If we do not need to "staple" NS data onto the record, this could simply be utf8 encoded Ucan

Additionally: the data could be CBOR-ified as well rather than a base64 encoding

rust/noosphere-ns/tests/ns_test.rs
jsantell: Cids are at least currently opaque to the entire name system --  that's changing a bit currently with the Ucan iteration, where Cids can be resolved in a Ucan signature (e.g. accessing a Store), so these tests can safely use `Blake2b256` form, good catch

rust/noosphere-ns/src/name_system.rs
jsantell: a Mutex seems like a heavy workaround for this, but there's probably some verbose simple solution as it seems to be a borrow checker "error"? (the hubris)

rust/noosphere-ns/src/records.rs
jsantell: New Ucan-based version does not have this issue, not sure what traits-bound hell I got myself into before

rust/noosphere-ns/src/name_system.rs
cdata: The `Mutex` may be a heavy workaround, but also: `get_record` being a mutating operation is counter-intuitive (although perfectly understandable when reading the implementation).

No action required, just food for thought.

rust/noosphere-ns/src/name_system.rs
cdata: (discussed offline, no action required)

rust/noosphere-ns/src/dht/processor.rs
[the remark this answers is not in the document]
jsantell: Oh i see, the error isn't in the log, nice catch!

rust/noosphere-ns/src/records.rs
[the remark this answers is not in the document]
cdata: Yah, I think even if JSON I would vote for using something ready to go out of https://github.com/subconsciousnetwork/noosphere/blob/main/rust/noosphere-storage/src/encoding.rs

I would prefer `CBOR`, but `DagJsonCodec` should also work with those methods (caveat we haven't used it anywhere).

jsantell:
Updated with UCAN integration, where records are authorized and set as UCAN tokens, and propagated on the network in their encoded form. Previous comments have been addressed; I'll inline the areas of interest from previous review -- `NSRecord` and `ns_test` are the majority of changes

rust/noosphere-ns/src/records.rs
```diff
@@ -0,0 +1,395 @@
...
+        assert_eq!(record.identity(), &sphere_identity);
+        assert_eq!(record.address(), Some(&Cid::from_str(cid_address).unwrap()));
+        record.validate(&store, &mut did_parser).await?;
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_nsrecord_delegated() -> Result<(), Error> {
+        // TODO
+        Ok(())
```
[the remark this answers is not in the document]
jsantell: What other scenarios can we test here that doesn't overlap with what the UCAN crate validates, or coordinating between multiple name systems (like in `ns_test.rs`)?

rust/noosphere-ns/src/records.rs
```diff
@@ -0,0 +1,395 @@
...
+
+        if self.address.is_none() {
+            return Err(anyhow!(
+                "Missing a valid fact entry with record address. {} {:?}",
+                identity,
+                self.token.facts()
+            ));
+        }
+
+        self.token.check_signature(did_parser).await?;
```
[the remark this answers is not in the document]
cdata: This may be useful: https://github.com/ucan-wg/rs-ucan/blob/9ef21415803ff4a64ff4c851adec160dcea2843c/ucan/src/ucan.rs#L92-L95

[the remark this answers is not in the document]
jsantell: That's used in `NSRecord::is_expired`, used during the "do i need to fetch from network" conditional in `NameSystem::set_record` --  `Ucan::is_expired` could be added to the general `NSRecord::validate` result, but then requires user/sphere interaction to occur at least within the original expiration date, even if several months, otherwise dead links could occur, the rational for *not* including it in the general validation. Although I haven't yet fleshed out the modality of "valid/authenticated" versus "authenticated but expired" and what the desired outcome of a once-active sphere looks like in the long tail, though maybe irrelevant within this context, as there's no gateway to propagate content/records

rust/noosphere-ns/src/records.rs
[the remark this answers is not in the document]
jsantell: Added tests for delegated capabilities

rust/noosphere-ns/src/records.rs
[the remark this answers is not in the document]
jsantell: Rolling expiration check into general validation based off of previous discussions, with the NS only propagating valid entries, but still enabling the possibility for e.g. sphere data to make decisions on a record based off of expiration date

rust/noosphere-ns/tests/ns_test.rs
```diff
@@ -0,0 +1,192 @@
...
+        .map_err(|e| anyhow!(e.to_string()))?
+        .pop()
+        .unwrap();
+    let bootstrap_addresses = vec![bootstrap_node.p2p_address().unwrap().to_owned()];
+
+    let store = SphereDb::new(&MemoryStorageProvider::default()).await?;
+    let mut name_systems: Vec<NSData> = vec![];
+
+    for _ in 0..ns_count {
+        let owner_sphere_key = generate_ed25519_key();
```
cdata, line 44: What does this key represent? We have been using `owner` to refer to a key that is a user's key. A `sphere`'s key is typically "burned" right after it is created (and doesn't directly sign or verify anything).

cdata (approved):
I don't have any feedback for changes, just some questions. This looks like really great progress, and we can land it when you are ready.

rust/noosphere-ns/tests/ns_test.rs
jsantell: Good Q: This key is used only for the DHT node, which is needed for libp2p implementation, as it derives `PeerId` (therefore the node's `Multiaddr` address) and all the underlying crypto/validation between peers. This represents the Gateway key. Though currently there is no connection "gateway key" and "DHT key" semantically, unless we added e.g. gateway keys to prioritize DHT records (#124). It could be an ephemeral key generated by the Gateway, though we lose static addressing in the network (i.e. "known good nodes" would be less useful if a gateway used new keys on every run)

rust/noosphere-ns/src/utils.rs
```diff
@@ -0,0 +1,49 @@
...
+            kind: Resource::Scoped(SphereReference {
+                did: sphere_did.to_owned(),
+            }),
+        },
+        can: SphereAction::Publish,
+    }
+}
+
+/// Generates a UCAN `"fct"` struct for the NS network, representing
+/// the resolved sphere's address as a [cid::Cid].
```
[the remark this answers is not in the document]
jsantell: :+1: good to nail down this verbiage throughout, I'll see if there's elsewhere that could be updated to "sphere's revision"

rust/noosphere-ns/Cargo.toml
```diff
@@ -24,15 +24,23 @@ readme = "README.md"
 [dependencies]
 anyhow = "^1"
 tracing = "0.1"
+cid = "~0.8"
+serde = "^1"
+serde_json = "^1"
+ucan = { version = "0.7.0-alpha.1" }
+ucan-key-support = { version = "0.7.0-alpha.1" }
+noosphere-storage = { version = "0.1.0-alpha.1", path = "../noosphere-storage" }
```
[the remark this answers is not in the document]
jsantell: :+1:

rust/noosphere-ns/tests/ns_test.rs
jsantell: Correction: The above refers to the `ns_key` -- the `owner_sphere_key` is the sphere the gateway is managing. With delegation working/testing, this could be updated to mimic that behavior

cdata (approved):
Great work as usual :raised_hands:

PR #123, opened 2022-11-02
feat(noosphere): Introduce `noosphere` crate

cdata:
The `noosphere` crate is the highest level entrypoint into Noosphere APIs that we offer. It re-exports the functionality of most of our other crates, and offers FFI interfaces for C and JS/WASM consumers.

In addition to the `noosphere` crate, this change introduces some major improvements to our Github Actions workflows. Workflows now let you build associated artifacts on demand, and are also assembled into a larger release orchestration.

NOTE: The `noosphere` crate is currently barebones because we just needed to get it to a point where we could compile it for different Apple targets and expose a trivial FFI. The focus of this change is on establishing our automation workflow for these kinds of artifacts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ target
 optimized_wasm
 dist
 .DS_Store
+noosphere.h
--- a/rust/noosphere/include/module.modulemap
+++ b/rust/noosphere/include/module.modulemap
@@ -1,0 +1,4 @@
+module Noosphere {
+    header "noosphere.h"
+    export *
+}
--- a/rust/noosphere/src/ffi.rs
+++ b/rust/noosphere/src/ffi.rs
@@ -1,0 +1,59 @@
+use std::collections::BTreeMap;
+
+use lazy_static::lazy_static;
+use safer_ffi::prelude::*;
+
+use crate::{
+    platform::{PlatformKeyMaterial, PlatformStore},
+    sphere::SphereContext,
+};
+
+lazy_static! {
+    static ref ACTIVE_SPHERES: BTreeMap<SphereHandle, Box<SphereContext<'static, PlatformKeyMaterial, PlatformStore>>> =
+        BTreeMap::new();
+}
+
+static mut GLOBAL_STORAGE_PATH: Option<String> = None;
+static mut SPHERE_STORAGE_PATH: Option<String> = None;
+
+ReprC! {
+    #[repr(C)]
+    pub struct SphereHandle {
+        user_identity: char_p::Box,
+        sphere_identity: char_p::Box,
+    }
+
+}
+
+ReprC! {
+    #[repr(C)]
+    pub struct SphereCreationResult {
+        mnemonic: char_p::Box
+    }
+}
+
+/// Set the global Noosphere storage path. This will be used to store user
+/// data and configuration that is cross-cutting with respect to sphere data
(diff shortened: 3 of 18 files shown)


rust/noosphere/src/ffi.rs
```diff
@@ -0,0 +1,59 @@
...
+        mnemonic: char_p::Box
+    }
+}
+
+/// Set the global Noosphere storage path. This will be used to store user
+/// data and configuration that is cross-cutting with respect to sphere data
+#[ffi_export]
+pub fn noosphere_set_global_storage_path(path: char_p::Ref<'_>) {
+    unsafe {
+        GLOBAL_STORAGE_PATH = Some(path.to_string());
```
jsantell, line 40: Not yet relevant for the scope of the FFI interface, but are rust's mutable statics thread safe?

rust/noosphere/src/lib.rs
```diff
@@ -0,0 +1,14 @@
...
+
+pub mod platform;
+pub mod sphere;
+// let key = create_key();
+// let sphere_context = noosphere_create_sphere(key);
+// noosphere_config_set(sphere_context, "gateway_url", "http://127.0.0.1:4433");
+// noosphere_
+
+// #[no_mangle]
+// pub unsafe extern "C" fn create_key()
```
jsantell, line 14: extra comments

rust/noosphere/src/sphere.rs
```diff
@@ -0,0 +1,69 @@
...
+where
+    K: KeyMaterial,
+    S: Store,
+{
+    Online {
+        client: Arc<Client<'a, K, SphereDb<S>>>,
+    },
+    Offline,
+}
+
```
jsantell, line 34: :eyes: excited to see where this interface goes

.gitignore
```diff
@@ -2,3 +2,4 @@ target
 optimized_wasm
 dist
 .DS_Store
+noosphere.h
```
jsantell, line 5: I wouldn't mind vendoring the header file once a ffi helper (safer-ffi, cbindgen, uniffi) if at least to link to the C API

rust/noosphere/include/module.modulemap
```diff
@@ -0,0 +1,4 @@
+module Noosphere {
+    header "noosphere.h"
+    export *
+}
```
jsantell, line 4: Is this for swift? How is this used?

jsantell (approved):
Great start on the FFI work!

rust/noosphere/src/ffi.rs
cdata: Not on their own, you need to use a `Mutex` or something with effectively the same guarantees.

.gitignore
cdata: We should make it trivial to generate it locally. I'm not opposed to checking it in, but it really should never be modified by hand.

rust/noosphere/include/module.modulemap
cdata: Yes, this is for XCode packaging ceremony. It is required for... reasons I don't fully understand, TBH. You need it to exist at the `.xcframework` generation step.

.gitignore
jsantell: +1

PR #130, opened 2022-11-08
dev: Expose replication/publication/ttl intervals to NameSystemBuilder

### LINKED ISSUE #127: NameSystem must propagate records within configured interval
The libp2p implementation implements [propagation and TTL configuration](https://docs.rs/libp2p/latest/libp2p/kad/struct.KademliaConfig.html#method.set_provider_publication_interval) for both value records and provider records. We should enable and test this.

jsantell:
Fixed #127

### DIFF
--- a/rust/noosphere-ns/src/builder.rs
+++ b/rust/noosphere-ns/src/builder.rs
@@ -39,7 +39,6 @@ where
     dht_config: DHTConfig,
     key_material: Option<Ed25519KeyMaterial>,
     store: Option<SphereDb<S>>,
-    propagation_interval: u64,
 }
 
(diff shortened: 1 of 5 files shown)


cdata (approved):
This looks good. I am assuming that these values weigh mostly on the behavior of the DHT, and are somewhat orthogonal to the time-based validity of UCANs in record values. Is that right?

jsantell:
> This looks good. I am assuming that these values weigh mostly on the behavior of the DHT, and are somewhat orthogonal to the time-based validity of UCANs in record values. Is that right?

Aye, unrelated to UCAN tokens' time-bounds

Issue #108, opened 2022-10-24
Consider type system alternative for stringly typed values


cdata:
We have a number of cases where parameters are string types: UCAN JWT strings, base64-encoded CIDs/signatures/whatever, DID strings.

The possibility that we might improve the legibility of these values came up in https://github.com/subconsciousnetwork/noosphere/pull/107#discussion_r1003804869. 

Let's discover the best approach to representing these stringly typed values in code and apply this strategy broadly in our code base. Type aliases and newtypes are worth considering, but there may be other alternatives.

jsantell:
TIL the name for single-valued types ([newtypes](https://doc.rust-lang.org/rust-by-example/generics/new_types.html)) -- this looks preferable to aliasing. Some considerations:

* How frequently do our stringly-typed values (STV) use string methods? (Naively, I'd guess "not often")
* Are there STVs that could benefit from other behaviours lifted into the newtype? e.g. `DID` having a `is_valid()` method
* Wonder if there's anything we can learn from [multiformats](https://github.com/multiformats), which admittedly are more complicated than our STVs I believe, but mostly provide additional context to some string-ifiable interface.

Some reasons for my preference in not using STVs:
* These simple STVs require a lot of context to interpret, leveraging layers of existing standards -- layers, as in, a single string represents a text-encoding, possibly other codec, semantic meaning e.g. "a DID", as well as possible variants (CIDv1 with Sha256BlakeSomething encoding)
* We re-document in many methods that "The identity is a utf-8 encoded DID string, representing the public key of a sphere" which could be elided into a newtype
* Stronger guarantees in wiring up different noosphere components (a DID string can't be passed to a function wanting a UCAN JWT string)

cdata:
> How frequently do our stringly-typed values (STV) use string methods? (Naively, I'd guess "not often")

We can make the newtypes effectively passthrough for this usecase by implementing `Deref<Target = String>` for them. For example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e72bf35655119246ede3af5fcc819d46

> Are there STVs that could benefit from other behaviours lifted into the newtype? e.g. DID having a is_valid() method

Probably! Especially as we delve into supporting other DID methods..

> Wonder if there's anything we can learn from [multiformats](https://github.com/multiformats), which admittedly are more complicated than our STVs I believe, but mostly provide additional context to some string-ifiable interface.

[...]

PR #131, opened 2022-11-09
feat: Add `noosphere` crate-based Swift package

cdata:
This change introduces a counterpart Swift package for the `noosphere` crate, in service of imminent integration between our iOS app and the Noosphere body of libraries. The `noosphere` crate compiles for both web and native targets. On native targets, it has an auxiliary FFI interface making it suitable for use from any language that can talk C ABI (such as Swift). The change also includes an `IndexedDB` + `WebCrypto` key storage backend, to ensure parity in the API on web.

At this time, it's capabilities are limited, but the intention is for us to relocate most of the substantive implementation of `noosphere-cli`'s commands under this crate, and re-implement `noosphere-cli` commands in terms of `noosphere` functionality.

Currently supported capabilities include: creating a key, and initializing a sphere.

Note that the `Package.swift` currently refers to a hand-uploaded artifact in a temporary Github repo that I intend to delete. This is necessary until we get a release of this library that uploads an artifact to an official Github release on the Noosphere repo, which can happen about as soon as this PR can be merged. At that time, we will update the `Package.swift` by hand to point to the new release.

I considered automating the `Package.swift` changes, but we can figure out if that would be useful down the road as we collectively hammer out our XCode workflow.

A word on FFI...

In Rust, it is very common practice to leverage macros to automatically generate bindings to other languages. A common approach for C ABI-speaking languages is to use a wonderful crate called `cbindgen`. The advantage of this approach is that it reduces a lot of code duplication. Instead of maintaining a C header file that matches, you can lean on codegen to produce the header for you.

However, one inescapable problem with FFI in Rust is that when you share access to memory with another language, you lose most of Rust's memory safety guarantees. So, FFI boundaries are almost always written in terms of `unsafe` Rust.

To partially alleviate this condition, we are using an alternative to `cbindgen` called [`safer-ffi`](https://github.com/getditto/safer_ffi). `safer-ffi` enables us to express our FFI interface in terms of __safe__ Rust (with some constraints on what can be expressed), and moves all of the `unsafe` parts into codegen. This strikes an appealing balance from a safety perspective: code that is maintained by humans is able to be checked for memory safety, and code that is `unsafe` can (in theory) be patched when there is a bug. It may turn out that `safer-ffi` is burdensome, but for now I would like us to give it a shot.

Here is the C header that is produced by `safer-ffi` as of this change:

```c
/*! \file */
/*******************************************
 *                                         *
 *  File auto-generated by `::safer_ffi`.  *
 *                                         *
 *  Do not manually edit this file.        *
 *                                         *
 *******************************************/

#ifndef __RUST_NOOSPHERE__
#define __RUST_NOOSPHERE__

#ifdef __cplusplus
extern "C" {
#endif

typedef struct NsSphereReceipt NsSphereReceipt_t;

/** \brief
 *  Read the sphere identity (a DID encoded as a UTF-8 string) from a
 *  [SphereReceipt]
 */
char const * ns_sphere_receipt_identity (
    NsSphereReceipt_t * const * sphere_receipt);

/** \brief
 *  Read the mnemonic from a [SphereReceipt]
 */
char const * ns_sphere_receipt_mnemonic (
    NsSphereReceipt_t * const * sphere_receipt);

/** \brief
 *  De-allocate a [SphereReceipt]
 */
void ns_sphere_receipt_free (
    NsSphereReceipt_t * sphere_receipt);

typedef struct NsNoosphereContext NsNoosphereContext_t;

/** \brief
 *  Initialize a brand new sphere, authorizing the given key to administer it.
 *  The returned value is a [SphereReceipt], containing the DID of the sphere
 *  and a human-readable mnemonic that can be used to rotate the key authorized
 *  to administer the sphere.
 */
NsSphereReceipt_t * ns_sphere_create (
    NsNoosphereContext_t * * noosphere,
    char const * owner_key_name);

/** \brief
 *  Join a sphere by initializing it and configuring it to use the specified
 *  key and authorization. The authorization should be provided in the form of
 *  a base64-encoded CID v1 string.
 */
void ns_sphere_join (
    NsNoosphereContext_t * * noosphere,
    char const * sphere_identity,
    char const * local_key_name,
    char const * authorization);

/** \brief
 *  Create a key with the given name in the current platform's support key
 *  storage mechanism.
 */
void ns_key_create (
    NsNoosphereContext_t const * noosphere,
    char const * name);

/** \brief
 *  Initialize a [NoosphereContext] and return a boxed pointer to it. This is
 *  the entrypoint to the Noosphere API, and the returned pointer is used to
 *  invoke almost all other API functions.
 * 
 *  In order to initialize the [NoosphereContext], you must provide two
 *  namespace strings: one for "global" Noosphere configuration, and another
 *  for sphere storage. Note that at this time "global" configuration is only
 *  used for insecure, on-disk key storage and we will probably deprecate such
 *  configuration at a future date.
 * 
 *  You can also initialize the [NoosphereContext] with an optional third
 *  argument: a URL string that refers to a Noosphere Gateway API somewhere
 *  on the network that one or more local spheres may have access to.
 */
NsNoosphereContext_t * ns_initialize (
    char const * global_storage_path,
    char const * sphere_storage_path,
    char const * gateway_url);

/** \brief
 *  De-allocate a [NoosphereContext]. Note that this will also drop every
 *  [SphereContext] that remains active within the [NoosphereContext].
 */
void ns_free (
    NsNoosphereContext_t * noosphere);


#ifdef __cplusplus
} /* extern "C" */
#endif

#endif /* __RUST_NOOSPHERE__ */
```

BREAKING CHANGE: The `noosphere-api` Client now holds an owned key instead of a reference.


rust/noosphere/src/sphere/context.rs
```diff
@@ -0,0 +1,133 @@
...
+
+/// A [SphereContext] is an accessor construct over locally replicated sphere
+/// data. It embodies both the storage layer that contains the sphere's data
+/// as the information needed to verify a user's intended level of access to
+/// it (e.g., local key material and [ucan::Ucan]-based authorization).
+/// Additionally, the [SphereContext] maintains a reference to an API [Client]
+/// that may be initialized as the network becomes available.
+///
+/// All interactions that pertain to a sphere, including reading or writing
+/// its contents and syncing with a gateway, flow through the [SphereContext].
```
jsantell, line 25: Great docs all around

rust/noosphere/src/ffi/mod.rs
```diff
@@ -0,0 +1,14 @@
+mod key;
```
jsantell, line 1: Consider: module level docs here indicating all(?) FFI representations must live in here

rust/noosphere/src/ffi/noosphere.rs
```diff
@@ -0,0 +1,74 @@
+use anyhow::Result;
+use safer_ffi::prelude::*;
+use url::Url;
+
+use crate::noosphere::{NoosphereContext as NoosphereContextImpl, NoosphereContextConfiguration};
+
+ReprC! {
+    #[ReprC::opaque]
+    pub struct NoosphereContext {
+        inner: NoosphereContextImpl
```
jsantell, line 10: Multiple e.g. `NoosphereContext` types seem like it could get confusing even with restricting the FFI interfaces to `noosphere::ffi::*` -- though I don't feel strongly, have no (good) alternatives, and the name collision is relatively contained.

rust/noosphere/src/ffi/noosphere.rs
```diff
@@ -0,0 +1,74 @@
...
+    }
```
jsantell, line 11: I like the FFI interface here (FFI opaque wrappers around their Rust counterparts)

Package.swift
```diff
@@ -0,0 +1,37 @@
+// swift-tools-version:5.5
```
jsantell, line 1: I think we discussed this, but there's no way (yet?) to move this to `swift/` yeah?

rust/noosphere/src/lib.rs
```diff
@@ -1,7 +1,14 @@
-// pub mod builder;
+#[macro_use]
```
jsantell, line 1: Consider: Adding docs at the crate level indicating how to generate new bindings (invisible unless looking at `examples/generate_header.rs`) -- or rolling into a `build.rs` builder that generates headers automatically

jsantell (approved):
Great work on wrangling FFIs in noosphere!

rust/noosphere/src/ffi/sphere.rs
```diff
@@ -0,0 +1,77 @@
...
+pub fn noosphere_free_sphere_receipt(sphere_receipt: repr_c::Box<SphereReceipt>) {
+    drop(sphere_receipt)
+}
+
+#[ffi_export]
+/// Initialize a brand new sphere, authorizing the given key to administer it.
+/// The returned value is a [SphereReceipt], containing the DID of the sphere
+/// and a human-readable mnemonic that can be used to rotate the key authorized
+/// to administer the sphere.
+pub fn noosphere_create_sphere(
```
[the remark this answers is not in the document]
cdata: Good idea, will fix!

rust/noosphere/src/noosphere.rs
```diff
@@ -0,0 +1,197 @@
...
+        let mut sphere_contexts = self.sphere_contexts.lock().await;
+        sphere_contexts.insert(sphere_identity.clone(), Arc::new(Mutex::new(context)));
+
+        Ok(SphereReceipt {
+            identity: sphere_identity,
+            mnemonic,
+        })
+    }
+
+    /// Join a sphere by DID identity, given a local key and an [Authorization]
```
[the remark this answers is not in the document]
cdata: The more I think about it, the more I want it to be a transparent newtype. Let's spitball some ideas in #108 .

jsantell:
Note: the FFI'd opaque structs are not namespaced:
```c
 SphereReceipt_t * noosphere_create_sphere (
    NoosphereContext_t * * noosphere,
    char const * owner_key_name);
```

While functions are (`noosphere_*`) the structs share global namespace in C, and for more generic structs, a collision could occur (not likely for `NoosphereContext`, but down the road a `Version` or `Storage` struct may). I'd say consider `NS*` prefix, though that's a rather overloaded prefix, and the C-style `ns_*` prefix (`ns_NoosphereContext`) looks tragic (maybe C-case for structs as well e.g. `ns_noosphere_context`?)

Also, for the C++ export, consider a namespace e.g. `namespace noosphere {}` when compiling as C++. Could also map from C-case `ns_noosphere_context` to C++ `noosphere::NoosphereContext` for context-appropriate casing. I know cbindgen has several flags for controlling these outputs, not sure about safer_ffi

rust/noosphere/src/platform.rs
```diff
@@ -1,26 +1,36 @@
 ///! Platform-specific types and bindings
+///! Platforms will vary in capabilities for things like block storage and
+///! secure key management. This module lays out the concrete strategies we will
+///! use on a per-platform basis.
 
 #[cfg(all(
```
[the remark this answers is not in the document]
cdata: This module is the "tip of the spear," so we are prescribing concrete implementations for each platform we support. Everywhere else we are generic by trait (although we have concrete platform implementations that are only compiled on the appropriate target).

Package.swift
cdata: No, the `Package.swift` has to be at the top-level.

jsantell (approved):

Issue #124, opened 2022-11-04
Providing fast NS lookups: evaluating provider records, querying strategies


jsantell:
The underlying DHT (utilizing the [libp2p Kademlia DHT spec](https://github.com/libp2p/specs/tree/master/kad-dht)) allows peers to query by a **TOPIC** for a **RESULT**, or query by **TOPIC** for all **PROVIDERS** that have **RESULT**. In the name system, we use a sphere's identity (DID) as a **TOPIC** and expect a UCAN auth token (or CID, TBD) as the **RESULT**, that can be verified and resolved into the sphere's latest revision (CID). DHT nodes handle these queries by sharing key/value records between peers. These records can be interpreted in two ways:

* Value Records: The value is the **RESULT**. Using Value Records is a push model, where the **RESULT** is propagated to peers.
* Provider Records: The value is a **PROVIDER** (specifically `PeerId` and `Multiaddr`) that provides **RESULT**. Using Provider Records is a pull model, where a node announces it has the **RESULT** for **TOPIC** by propagating that record to peers. Nodes querying in this model query for a list of providers of this record (discovery), and can subsequently dial each other. The process of retrieving **RESULT** from **PROVIDER** is outside of the scope of the DHT's spec, though the rust-libp2p implementation allows us to hook into this via e.g. `Bitswap`

----

Currently, the name system uses Value Records for simplicity as we shape out how the system works. The DHT **RESULT** is a UCAN token verifying a sphere's key delegate with "sphere/publish" capabilities, along with the final value the name system is responsible for, the sphere's revision address. When a new revision is published, the gateway/NS hosts and sends the new record to its peers, and in intervals to possibly different peers. Currently in the v1 name system, querying the network for results, we grab the first valid record we can find (which in small networks, may be already on the node) -- though there may be many valid/verifiable records for a single identity within TTL. We may want several options during lookups specifying just wanting the first record found (fast!), versus waiting for 3 or more and taking the latest. There's no guarantee that there's not a newer revision that has not yet been discovered in either case.

Using provider records instead, as pragmatically only a sphere's corresponding gateway would be propagating records, could mean a single **PROVIDER** for a **TOPIC**. As the actual **RESULT** is pulled on-demand, the gateway with fresh records will always be the found peer, e.g. fresh results when querying outside of the local cache. This does of course require the gateway to be always available, and still possible for other nodes to announce themselves as providers for the same records a gateway is hosting.

[...]

PR #138, opened 2022-11-10
docs: Initial write up of the Noosphere Name System spec.

jsantell:
Initial draft here. Goal here is to define all properties used in the system in order for an independent client to participate in the Noosphere Name System. #124 may change this to be Provider-based, and there's some improvements/clarifications to be made here, but wanted to start discussion.

### DIFF
--- a/design/name-system.md
+++ b/design/name-system.md
@@ -1,0 +1,113 @@
+# Noosphere Name System
+
+The Noosphere Name System (NNS) is a distributed network, similar to [DNS], that resolves a sphere's identity to the latest address of a revision of the sphere.
+
+## :warning: Status: Experimental :warning:
+
+Noosphere is under active development and we are still working through some challenges, and anything documented here can change without warning during this time. This specification will be used in the future to support new clients using the Noosphere Name System for Noosphere, or possibly other related data, but for now should be considered internal documentation.
+
+## Terms
+
+* **Spheres** are a core structure in Noosphere representing an entire "workspace" or "notebook", containing many documents, with change history represented as a series of revisions, not dissimilar from Git. A sphere's capabilities are managed by a public/private [Ed25519] encryption key. The sphere's unique identity is represented and publically referenced as a [DID]. 
+* **Decentralized Identifiers** ([DID]s) are public keys of a public/private key pair represented as a text string. The DIDs used in NNS are all derived from [Ed25519] encryption keys.
+  * Example representation: `did:key:z6MkoE19WHXJzpLqkxbGP7uXdJX38sWZNUWwyjcuCmjhPpUP`
+* **User Controlled Authorization Networks** ([UCAN]s) are an extension of JSON Web Tokens ([JWT], [RFC7519]), designed to enable ways of authorizing offline-first apps and distributed systems. Represented as JSON data, base64 encoded as a string.
+* **[Multiaddr]s** is an extensible format describing a network address through multiple layers and protocols.
+  * Example representation for a machine at IP address 10.20.30.1 with TCP port 20000: `/ip4/10.20.30.1/tcp/20000`
+* **[PeerId]s** are [multihash]ed public keys. In NNS, PeerIds are derived from keys unrelated to spheres, and used to identify and communicate with DHT nodes in the network.
+* **Content Identifiers** ([CID]s) are [multihash]ed [content addresses](https://docs.ipfs.tech/concepts/content-addressing/), where content can be identified by an immutable address. CIDs used in NNS are all CIDv1 hashed via [BLAKE2b256].
+
+## Overview
+
+A sphere in Noosphere contains internal data for an address book, where other sphere's identities ([DID]s) can be stored. DIDs are hardly memorizable, so they're stored with a name, like "Alice". When Bob wants to reference Alice's notes on dogs from his sphere in subtext, it'd look like `"@Alice/dogs"`. The "Alice" entry in the address book maps to a DID, like `did:key:z6MkoE19WHXJzpLqkxbGP7uXdJX38sWZNUWwyjcuCmjhPpUP`, but a [CID] address is needed in order to fetch, ideally the latest, contents from Alice's sphere. This is where the Noosphere Name System comes in.
+
+Bob's gateway connects to the NNS, asking peers if anyone has an address for Alice's key. The address can be found from the network, and using [UCAN], can be validated that it was Alice publishing it from Alice's gateway, or maybe even from another peer on the network that has no idea who Alice is. Once the address is retrieved, Alice's sphere's content can be fetched and remixed.
+
+Later, Bob wants Alice's latest content. Because these addresses are content-addressed CIDs, they're immutable, meaning everytime a sphere's content changes, it has a new address. Bob's gateway connects to the NNS again, looking for a newer address for Alice's sphere, comparing the UCAN timestamps from previous and current results for Alice's sphere.
+
+Additionally, Bob's gateway is propagating Bob's sphere's latest revision address when the sphere changes for others to find.
+
+## Network Spec
+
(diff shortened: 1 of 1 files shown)


cdata:
This is awesome :tada: I can't wait to dive in..

design/name-system.md
```diff
@@ -0,0 +1,103 @@
...
+* **Decentralized Identifiers** ([DID]s) are public keys of a public/private key pair represented as a text string. The DIDs used in NNS are all derived from [Ed25519] encryption keys.
+  * Example representation: `did:key:z6MkoE19WHXJzpLqkxbGP7uXdJX38sWZNUWwyjcuCmjhPpUP`
+* **User Controlled Authorization Networks** ([UCAN]s) are an extension of JSON Web Tokens ([JWT], [RFC7519]), designed to enable ways of authorizing offline-first apps and distributed systems. Represented as JSON data, base64 encoded as a string.
+* **[Multiaddr]s** is an extensible format describing a network address through multiple layers and protocols.
+  * Example representation for a machine at IP address 10.20.30.1 with TCP port 20000: `/ip4/10.20.30.1/tcp/20000`
+* **[PeerId]s** are [multihash]ed public keys. In NNS, PeerIds are derived from keys unrelated to spheres, and used to identify and communicate with DHT nodes in the network.
+* **Content Identifiers** ([CID]s) are [multihash]ed [content addresses](https://docs.ipfs.tech/concepts/content-addressing/), where content can be identified by an immutable address. CIDs used in NNS are all CIDv1.
```
jsantell, line 14: Should we specify any specific hash type for CIDv1?

design/name-system.md
```diff
@@ -0,0 +1,103 @@
+# Noosphere Name System
+
+The Noosphere Name System (NNS) is a distributed network, similar to [DNS], that resolves a sphere's identity to the latest address of a revision of the sphere.
+
+## Terms
+
+* **Spheres** are a core structure in Noosphere representing an entire "workspace" or "notebook", containing many documents, with change history represented as a series of revisions, not dissimilar from Git. A sphere's capabilities are managed by a public/private [Ed25519] encryption key. The sphere's unique identity is represented and publically referenced as a [DID].
```
jsantell, line 7: These terms can be expanded into the general glossary as well

jsantell (commented):
Pedantic feedback welcome! There are a lot of encodings/hashes/formats/ways to interpret bytes here

design/name-system.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Protocols
+
+The DHT nodes **MUST** implement the libp2p [Identify] spec.
+
+### DHT Methods 
+
+All methods in the underlying spec **MUST** be implemented, with cavaets below.
+
+The network uses Value Records (using `GET_VALUE`, `PUT_VALUE` DHT methods) where the key **SHOULD** be a sphere's public [Ed25519] key encoded as a [DID] UTF-8 string, and the corresponding value **SHOULD** be a [UCAN] token encoded as a [JWT] string.
```
jsantell, line 42: The value could be a CID itself in the future, resolving into a UCAN. Should be straight forward to implement this. I think defining both forms is reasonable

cdata:
@jsantell great write up.

At this early stage, it might be nice to err on the side of "this is our intended design," and further away from, "this is a spec for other people to implement."

The reason why is: we do not yet know ourselves if our intended design will survive the harsh gauntlet of practical deployment.

Some other feedback:

It would be nice to see those FAQs answered, or else to move them into issues if you don't think we can answer them now.
We have had some discussion in the past about provider records vs value records, and you opened https://github.com/subconsciousnetwork/noosphere/issues/124 to talk about this a bit. To the extent that we may need to come to some conclusion related to discussion in  #124, perhaps we can add a TODO in this document.

design/name-system.md
```diff
@@ -0,0 +1,103 @@
...
+  "aud": "did:key:z6MkkVfktAC5rVNRmmTjkKPapT3bAyVkYH8ZVCF1UBNUfazp",
+  "att": [{
+    "with": "sphere:did:key:z6MkkVfktAC5rVNRmmTjkKPapT3bAyVkYH8ZVCF1UBNUfazp",
+    "can": "sphere/publish"
+  }],
+  "prf": [
+    /* NEED TO ADD PROOFS */
+  ],
+  "fct": [{
+    "address": "bafy2bzacec4p5h37mjk2n6qi6zukwyzkruebvwdzqpdxzutu4sgoiuhqwne72"
```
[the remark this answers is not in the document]
cdata: "link" might hit the sweet spot. It's the IPLD domain term for a CID in the data structure.

I would vote to start small, and fall back to observing the UCAN spec (which allows for arbitrary facts).

design/name-system.md
cdata: Although we discussed this possibility, worth noting that there is a likely performance advantage to making it a JWT and not a CID pointing to a JWT: the top-level record is likely to change a lot, but its associated proofs are likely to stay the same for long time. So, by making the record a JWT, we always send around the part that changes every time, and can rely on caching to take care of the parts that don't change much.

design/name-system.md
cdata: We're using Blake2b256 for everything right now. We may not be able to get away with requiring a specific hash over the long run, but for now it saves us time to assume this hash everywhere.

design/name-system.md
```diff
@@ -0,0 +1,103 @@
...
+All methods in the underlying spec **MUST** be implemented, with cavaets below.
+
+The network uses Value Records (using `GET_VALUE`, `PUT_VALUE` DHT methods) where the key **SHOULD** be a sphere's public [Ed25519] key encoded as a [DID] UTF-8 string, and the corresponding value **SHOULD** be a [UCAN] token encoded as a [JWT] string.
+
+* `PUT_VALUE` methods **SHOULD** reject malformed UCAN tokens or UCAN tokens that do not pass Validation](#validation).
+* `PUT_VALUE` and `GET_VALUE` methods with non-DID keys **SHOULD** be ignored.
+* `ADD_PROVIDER` and `GET_PROVIDERS` methods are not currently used and **MAY** be fulfilled.
+* `FIND_NODE` and `PING` **MUST** passthrough from the underlying spec.
+
+### UCAN SpherePublish Token
```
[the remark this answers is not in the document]
jsantell: Aye, WDYT:

Could be removed, defer to spec:
* "Issuer"
* "Facts"
* "Proofs"

Context-specific:
* "Audience": used in proper UCAN form, but I think it's relevant to specify what identity is the "target" (e.g. the sphere associated with the Record)
* "Attenuation": the mapping of the RESOURCE to AUDIENCE with token capability "sphere/publish"
* "Facts": the mapping of token "address" to CID

design/name-system.md
jsantell: Perfect, will treat similarly as Ed25519-based DID keys (we use them, others may be supported but TBD).

design/name-system.md
jsantell: great points, will note that!

jsantell:
> @jsantell great write up.
> 
> At this early stage, it might be nice to err on the side of "this is our intended design," and further away from, "this is a spec for other people to implement."
> 
> The reason why is: we do not yet know ourselves if our intended design will survive the harsh gauntlet of practical deployment.
> 
> Some other feedback:
> 
> It would be nice to see those FAQs answered, or else to move them into issues if you don't think we can answer them now. We have had some discussion in the past about provider records vs value records, and you opened #124 to talk about this a bit. To the extent that we may need to come to some conclusion related to discussion in #124, perhaps we can add a TODO in this document.

:100: agreed, the "experimental" status and contract, and TODOs should be completed before landing (more time-bound than clarity-bound). We have answers on the alternatives, and should be more explicit about the possibility (which I find likely) of using provider records

jsantell:
Added BLAKE2b256 references, mention experimental status, fill out some FAQs, and cut out the base-UCAN explained-fields

PR #139, opened 2022-11-10
chore: cleanups on some older DHT code and docs

jsantell:
Some cleanups from earlier DHT work, now with a clearer understanding of the system as a whole (and tidying up docs)

* Rename `set_record` terms to `put_record` to align with the underlying DHT methods -- similarly, use `key` instead of `name` for records.
* Use `&[u8]` interfaces for the DHT module
* `#[cfg(doc)]` some modules to cleanup references in docs
* Simplify DHT tests (thanks to above changes) with some util cleanups so we can start to splice together stranger network in tests.


cdata (approved):
Looks like a solid cleanup :raised_hands:

PR #140, opened 2022-11-11
feat: `SphereFs` is initialized with key material

### LINKED ISSUE #132: `SphereFs` should receive credentials during initialization
While implementing #131 , I realized that for ergonomic reasons, we probably want to initialize `SphereFs` with a key and authorization. This is a breaking change from the current API, which asks for key and authorization at "save" time.

There are at least two reasons why we want this:

 1. In practice we will want content to be signed as well as the sphere revision; currently only the sphere revision is signed, and we would need to ask for credentials at "write" time if we wanted content to be signed as well
 2. In the case of high-level APIs like the one in the `noosphere` crate, we can avoid asking the API user to configure credentials multiple times

cdata:
This change is primarily intended to move `SphereFs` towards a model where key material and authorization are provided up-front. This enables significantly more ergonomic use, where the author can provide their credentials once and then write-to and save the sphere as many times as they want. It also will lay a path for automatic signing of content as it is written (where currently we only sign the sphere as it is saved).

Along the way to implementing this change, we have consolidated the pairing of a `KeyMaterial` and an `Authorization` into a single construct called an `Author`. This reduces the complexity of many of our APIs (which all asked for a pairing of these constructs).

While we were doing that, I went ahead and ripped off a band-aide related to #108 . A lot of our APIs have been transitioned towards expecting a `Did` where previously they expected a string that was implicitly documented to be a DID string. This should offer greater explicitness in our signatures and reduce the burden of documenting what DIDs are all over the place.

BREAKING CHANGE: Many APIs that previously asked for bare strings when a DID string was expected now expect a newtype called `Did` that wraps a string.

BREAKING CHANGE: APIs that previously expected both a `KeyMaterial` and an `Authorization` now expect a consolidated construct called an `Author` that pairs these two things.

Fixes #132


rust/noosphere-core/src/authority/author.rs
```diff
@@ -0,0 +1,163 @@
...
+    ReadWrite,
+    ReadOnly,
+}
+
+/// An author is a user or program who is reading content from and/or writing
+/// content to a sphere. This construct collects the identity and the
+/// authorization of that entity to make it easier to determine their level of
+/// access to the content of a given sphere.
+#[derive(Clone)]
+pub struct Author<K>
```
jsantell, line 31: :+1:

rust/noosphere-cli/src/native/workspace.rs
```diff
@@ -201,7 +201,7 @@ impl Workspace {
 
         let content = self.read_local_content(pattern, new_blocks).await?;
 
-        let sphere_fs = SphereFs::at(&sphere_did, &sphere_cid, None, db);
+        let sphere_fs = SphereFs::at(&sphere_did, &sphere_cid, &Author::anonymous(), db).await?;
```
jsantell, line 204: I don't quite understand the meaning for an anonymous (new key) author here, could you explain?

jsantell (approved):

rust/noosphere-cli/src/native/workspace.rs
cdata: When you write to a sphere, you need a DID to be used as the author of the change; this brings a need to always have a DID that corresponds to the author. Eventually, this `SphereFs` construct will let you bounce from one sphere to the next by address book, moving your credentials along with it. In some spheres, your credential will give you read/write, in others read-only. An "anonymous" key is guaranteed to always be read-only.

A cheaper alternative would be to design the author to have some dummy DID that resolves to nothing, but generating a key is fairly cheap to do and would spare us from weird cases where someone tries to resolve a dummy DID for some reason.

PR #143, opened 2022-11-14
dev: First pass on a `bootstrap` CLI

jsantell:
dev: First pass on a `bootstrap` CLI in noosphere-ns to spin up DHT bootstrap nodes for the name system.

### DIFF
--- a/rust/noosphere-ns/Cargo.toml
+++ b/rust/noosphere-ns/Cargo.toml
@@ -26,6 +26,7 @@ readme = "README.md"
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 anyhow = "^1"
 tracing = "0.1"
+lazy_static = "^1"
 cid = "~0.8"
 serde = "^1"
 serde_json = "^1"
@@ -37,9 +38,25 @@ tokio = { version = "1.15", features = ["io-util", "io-std", "sync", "macros", "
 libp2p = { version = "0.49.0", default-features = false, features = [ "identify", "dns", "tcp", "tokio", "noise", "mplex", "yamux", "kad" ] }
 noosphere-storage = { version = "0.1.0", path = "../noosphere-storage" }
 noosphere-core = { version = "0.2.0", path = "../noosphere-core" }
+noosphere = { version = "0.2.0", path = "../noosphere", optional = true }
+clap = { version = "^4", features = ["derive"], optional = true }
+home = { version = "~0.5", optional = true }
+toml = { version = "~0.5", optional = true }
 
 [dev-dependencies]
+
+[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
 rand = { version = "0.8.5" }
 test-log = { version = "0.2.11", default-features = false, features = ["trace"] }
 tracing-subscriber = { version = "~0.3", features = ["env-filter"] }
 libipld-cbor = "~0.14"
+tempdir = { version = "~0.3" }
+
+[features]
+default = ["bootstrap"]
+bootstrap = ["clap", "noosphere", "home", "toml"]
+
+
+[[bin]]
+name = "bootstrap-nns"
+required-features = ["bootstrap"]
--- a/rust/noosphere-ns/README.md
+++ b/rust/noosphere-ns/README.md
@@ -3,3 +3,12 @@
 # noosphere-ns
 
 Noosphere's P2P name system.
+
+## Bootstrap Node
+
+The `bootstrap_nns` binary target is an executable that runs one or many bootstrap
+nodes, based on configuration.
+
+```
+cargo run --bin bootstrap_nns -- run --key my-key-name --port 6666
+```
--- a/rust/noosphere-ns/src/name_system.rs
+++ b/rust/noosphere-ns/src/name_system.rs
@@ -17,6 +17,15 @@ use crate::NameSystemBuilder;
 #[cfg(doc)]
 use cid::Cid;
 
+pub static BOOTSTRAP_PEERS_ADDRESSES: [&str; 1] =
+    ["/ip4/134.122.20.28/tcp/6666/p2p/12D3KooWAKxaCWsSGauqhCZXyeYjDSQ6jna2SmVhLn4J7uQFdvot"];
+
+lazy_static! {
+    /// Noosphere Name System's maintained list of peers to
+    /// bootstrap nodes joining the network.
+    pub static ref BOOTSTRAP_PEERS: [Multiaddr; 1] = BOOTSTRAP_PEERS_ADDRESSES.map(|addr| addr.parse().expect("parseable"));
+}
+
 /// The [NameSystem] is responsible for both propagating and resolving Sphere DIDs
 /// into an authorized UCAN publish token, resolving into a [Cid] address for
 /// a sphere's content. These records are propagated and resolved via the
@@ -252,5 +261,17 @@ where
         if let Err(e) = self.disconnect() {
             error!("{}", e.to_string());
         }
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn bootstrap_peers_parseable() {
+        // Force getting the lazy static ensuring the addresses
+        // are valid Multiaddrs.
+        assert_eq!(BOOTSTRAP_PEERS.len(), 1);
     }
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1033,6 +1033,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "fuchsia-cprng"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a06f77d526c1a601b7c4cdd98f54b5eaabffc14d5f2f0296febdc7f357c6d3ba"
+
+[[package]]
 name = "futf"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2471,16 +2477,22 @@ dependencies = [
  "anyhow",
  "async-trait",
  "cid",
+ "clap",
  "futures",
+ "home",
+ "lazy_static",
  "libipld-cbor",
  "libp2p",
+ "noosphere",
  "noosphere-core",
  "noosphere-storage",
  "rand 0.8.5",
  "serde",
  "serde_json",
+ "tempdir",
  "test-log",
  "tokio",
+ "toml",
  "tracing",
  "tracing-subscriber",
  "ucan",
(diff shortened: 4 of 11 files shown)


rust/noosphere-ns/Cargo.toml
```diff
@@ -37,9 +38,21 @@ tokio = { version = "1.15", features = ["io-util", "io-std", "sync", "macros", "
 libp2p = { version = "0.49.0", default-features = false, features = [ "identify", "dns", "tcp", "tokio", "noise", "mplex", "yamux", "kad" ] }
 noosphere-storage = { version = "0.1.0", path = "../noosphere-storage" }
 noosphere-core = { version = "0.1.0", path = "../noosphere-core" }
+noosphere = { version = "0.1.0-alpha.1", path = "../noosphere", optional = true }
```
cdata, line 41: :sunglasses:

rust/noosphere-ns/src/name_system.rs
```diff
@@ -17,6 +17,15 @@ use crate::NameSystemBuilder;
 #[cfg(doc)]
 use cid::Cid;
 
+pub static BOOTSTRAP_PEERS_ADDRESSES: [&str; 1] =
+    ["/ip4/134.122.20.28/tcp/6666/p2p/12D3KooWAKxaCWsSGauqhCZXyeYjDSQ6jna2SmVhLn4J7uQFdvot"];
```
cdata, line 21: Do you think in time it may make sense to accept this list via environment or CLI configuration?

cdata (dismissed):
Some incidental questions for my own enrichment, but this looks like a great start :boat:

rust/noosphere-ns/Cargo.toml
```diff
@@ -37,9 +38,21 @@ tokio = { version = "1.15", features = ["io-util", "io-std", "sync", "macros", "
...
 rand = { version = "0.8.5" }
 test-log = { version = "0.2.11", default-features = false, features = ["trace"] }
 tracing-subscriber = { version = "~0.3", features = ["env-filter"] }
 libipld-cbor = "~0.14"
+
+[features]
+bootstrap = ["clap", "noosphere", "home"]
+
+[[bin]]
+name = "bootstrap"
```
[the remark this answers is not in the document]
jsantell: :+1:

rust/noosphere-ns/src/name_system.rs
jsantell: These should be rather permanent and ubiquitous, and should be the default list of peers for most clients -- that being said, the name system (and individual DHT nodes) do not prescribe this list, and this bootstrap script passes this list along (filtering out its own address, some work needs to be done on wrangling the multiple forms of addresses here, and how they're referenced etc.).

Specifically on *customizing bootstrap peers via CLI/env*, that could be useful for the "run your own bootstrap nodes" flow, though no immediate need yet

rust/noosphere-ns/src/bin/bootstrap_nns/cli.rs
```diff
@@ -0,0 +1,50 @@
...
+#[clap(name = "bootstrap_nns")]
+pub struct CLI {
+    #[clap(subcommand)]
+    pub command: CLICommand,
+}
+
+#[derive(Subcommand)]
+pub enum CLICommand {
+    Install,
+    Uninstall,
```
jsantell, line 17: Looking into systemd daemons that we can un/register via this CLI for our VPS.

rust/noosphere-ns/src/bin/bootstrap_nns/main.rs
```diff
@@ -0,0 +1,38 @@
...
+        _ => {
+            return Err(anyhow!("Unimplemented."));
+        }
+    };
+
+    // Allow aborting (ctrl+c) during the initial run,
+    // and after (when we want to wait exclusively for ctrl+c signal)
+    let mut aborted = false;
+    tokio::select! {
+        _ = tokio::signal::ctrl_c() => { aborted = true; },
```
jsantell, line 30: Maybe there's a better way here, e.g. have `runner.run()` sleep the thread after setting things up. Would like to avoid signaling/messaging INTO runner to abort, unless there's a need for it

rust/noosphere-ns/src/bin/bootstrap_nns/cli.rs
```diff
@@ -0,0 +1,50 @@
...
+        key: Option<String>,
+
+        /// If no configuration path provided, the listening port of this DHT node.
+        #[clap(short, long)]
+        port: Option<u16>,
+
+        /// If no configuration path provided, a flag to create key with name `--key`
+        /// if one does not already exist.
+        #[clap(long)]
+        create_key_if_needed: bool,
```
jsantell, line 37: `create_key_if_needed` is a temporary solution to provide some key gen capabilities here without pulling in `orb`. Open to suggestions here, I'm mostly indifferent, wanting to keep it simple (as opposed to a whole `key` subcommand)

rust/noosphere-ns/Cargo.toml
```diff
@@ -37,9 +38,27 @@ tokio = { version = "1.15", features = ["io-util", "io-std", "sync", "macros", "
...
 tracing-subscriber = { version = "~0.3", features = ["env-filter"] }
 libipld-cbor = "~0.14"
+tempdir = { version = "~0.3" }
+
+[features]
+default = []
+bootstrap = ["clap", "noosphere", "home", "toml"]
+
+[target.'cfg(not(target_arch = "wasm32"))'.features]
+default = ["bootstrap"]
```
jsantell, line 60: rationale for this being a default feature is integration into larger test suite. Otherwise we'd need to flag --features="bootstrap" everywhere, preferring for library usage to disallow this. WDYT?

Separately, this is cfg-flagged because otherwise the web builds try to build the binary, even with the main file having `#![cfg(not(target_arch = "wasm32"))]`

rust/noosphere-ns/src/bin/bootstrap_nns/cli.rs
cdata: In the interest of separating concerns, is it possible to set up a daemon separately from the CLI implementation? Or does there have to be a special code path when running as a daemon?

rust/noosphere-ns/Cargo.toml
cdata: As with the CLI, I think you may end up being happier allowing this to be built for `wasm32-unknown-unknown` but stubbing out the main in that case. The reason is: you will probably want to test your underlying library, so you have to stub that part out _anyway_.

rust/noosphere-ns/README.md
```diff
@@ -3,3 +3,12 @@
...
 
 Noosphere's P2P name system.
+
+## Bootstrap Node
+
+The `bootstrap_nns` binary target is an executable that runs one or many bootstrap
+nodes, based on configuration.
+
+```
+cargo run --bin bootstrap_nns -- run --key my-key-name --port 6666
```
cdata, line 13: Any particular reason to use `_` instead of `-`? Most commands I know that are separated like this use `-`...

rust/noosphere-ns/src/bin/bootstrap_nns/cli.rs
cdata: In the CLI and in our Rust APIs (including FFI), we put the burden on the user to create a key. This makes more sense in those cases: for users, there is usually a user-given ID (e.g., a username) given that the key should be associated with so that when they perform a future TouchID authorization (or equivalent action), they see the name of the key somewhere in the UI.

I'm not as opinionated for the bootstrap nodes, but if you want to be consistent with the CLI then the key should just be assumed to be created in advance (or else it's an error).

rust/noosphere-ns/src/bin/bootstrap_nns/cli.rs
jsantell: Could move it to our "devops scripts", as we mostly just need to drop a text config file pointing at this executable and run a systemd command, so ya

jsantell: Will remove these stubs -- I think I'll leave `run` as a subcommand, as there a handful of other things we can do here that'd be valuable (querying, maybe keygen stuff, if kicking daemon-ish stuff to systemd then maybe there aren't many more commands here)

rust/noosphere-ns/src/bin/bootstrap_nns/main.rs
```diff
@@ -0,0 +1,38 @@
...
+use runner::{Runner, RunnerConfig};
+use tokio;
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    let key_storage = InsecureKeyStorage::new(&utils::get_keys_dir()?)?;
+
+    let mut runner = match CLI::parse().command {
+        command @ CLICommand::Run { .. } => {
+            Runner::from(RunnerConfig::try_from_command(&key_storage, command).await?)
```
cdata, line 19: What does `runner` do?

rust/noosphere-ns/Cargo.toml
jsantell: > The reason is: you will probably want to test your underlying library, so you have to stub that part out anyway.

as in, test the bootstrap library on wasm32-unknown-unknown?

(I'll take a look at the CLI's cargo config)

rust/noosphere-ns/src/name_system.rs
cdata: Acknowledging that in the typical, these peers should be long-lived and well-known, I think we may need the flexibility for this list to be customizable. Example use cases:

 - I have private infrastructure and don't want my nodes to bootstrap off of publicly available peers
 - I'm an infrastructure provider and my well-known peers are a dynamic detail of my cloud infrastructural configuration

No action required for this change though.

rust/noosphere-ns/README.md
jsantell: :+1: aye `-` makes more sense and is consistent with pretty much everything in `/usr/bin/`

rust/noosphere-ns/Cargo.toml
cdata: No, I mean: no matter what you are going to have to deal with this command running:

```sh
cargo run --target wasm32-unknown-unknown
```

Which means even if you gate the _binary_ on a feature, you still have tests invoking the library code behind that binary (which will have to be conditioned not to run on `wasm32` arch anyway).

jsantell (commented):
Latest version cleans up the wasm32 conditionals, removes the keygen fallback (instead for now, is its own command, could defer to other mechanisms, orb, rsync, no preference), addressed comments -- adding bootstrap peers in a follow up

cdata (approved):
Great start for this stand-alone binary!

PR #154, opened 2022-11-15
Introduce pet names to spheres

### LINKED ISSUE #153: Implement pet name "address book" for `SphereIpld` and related constructs
The current iteration of the `SphereIpld` data structure is missing a `VersionedMapIpld` field that records the pet names and DIDs of other spheres the user is "following." We should extend the data structure to incorporate this information.

Here is the suggested approach:

We will define a data structure for records in the "address book":

```rs
// In noosphere-core/data:
struct SphereAddressIpld {
  identity: Did, // Identity of the sphere
  last_record: Option<UcanIpld> // Last value resolved from the name system, if any
}
```

And a `names` field will be added to `SphereIpld` with the signature:

```rs
// The key is the user-asigned pet name of the sphere
pub type Names = VersionedMapIpld<String, SphereAddressIpld>
```

Relevant to #12 and #49

cdata:
This is a small-ish change to get our feet in the water with pet names. It introduces a new `VersionedMap` for pet names, where we are mapping a human-readable name to what we call an `AddressIpld`. The address records a DID, and optionally a last known resolved value (which is represented as a JWT).

I also threw in some cleanup: I wrote a declarative macro for enhancing a string-wrapper struct with trait implementations to make the wrappers coherent with strings, and then added string-wrapper structs for JWTs and BIP-39 mnemonics. I also started propagating these structs a bit more.

Fixes #153

### DIFF
--- a/rust/noosphere-core/src/data/strings.rs
+++ b/rust/noosphere-core/src/data/strings.rs
@@ -2,6 +2,62 @@ use std::{fmt::Display, ops::Deref};
 
 use serde::{Deserialize, Serialize};
 
+/// A helper to stamp out trait implementations that promote coherence between
+/// Rust strings and a given wrapper type
+macro_rules! string_coherent {
+    ($wrapper:ty) => {
+        impl Deref for $wrapper {
+            type Target = String;
+
+            fn deref(&self) -> &Self::Target {
+                &self.0
+            }
+        }
+
+        impl From<&str> for $wrapper {
+            fn from(value: &str) -> Self {
+                Self(value.to_owned())
+            }
+        }
+
+        impl From<String> for $wrapper {
+            fn from(value: String) -> Self {
+                Self(value)
+            }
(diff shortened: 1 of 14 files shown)


rust/noosphere-core/src/data/strings.rs
```diff
@@ -1,7 +1,72 @@
-use std::{fmt::Display, ops::Deref};
+use std::{fmt::Display, hash::Hash, ops::Deref};
 
 use serde::{Deserialize, Serialize};
 
+/// A helper to stamp out trait implementations that promote coherence between
+/// Rust strings and a given wrapper type
+macro_rules! string_coherent {
```
jsantell, line 7: very nice

jsantell (dismissed):
Looks great! Looks like there's some CI errors from missing files

jsantell (approved):

PR #159, opened 2022-11-16
chore: fix link typo in name system spec, rename name system binary

jsantell:
Some cleanups from bootstrap script. Renaming the binary utility to `nns`, as "bootstrap" is what we specifically do to enable discovery in a noosphere name system network, and allows additional methods (e.g. "query") that operate on a noosphere name system, unrelated to the act of "bootstrapping".


cdata:
I'm totally fine with `nns` (and I think it's probably the most appropriate name for this) but I want to call out for posterity that I heard @blaine use this name for Fission's aspirational name system before we landed this, so they should get dibs on it if they want it :sweat_smile:

cdata (approved):

blaine:
Thanks @cdata! My current plan is to call the protocol NNS, and since a single executable should be all that's needed to join/run/do lookups against the service, nns feels like an appropriate executable name for it, if it's not too much trouble to avoid a conflict for now? I've made some progress on the design, writeup soon. This week, though... what even is this week. 🙃

(haven't had a chance to review your code here again, just playing catch-up, but I could imagine this actually being the basis for the nns service in any event! I have no precious notions of us writing code from scratch vs contributing to your start here)

jsantell:
> Thanks @cdata! My current plan is to call the protocol NNS, and since a single executable should be all that's needed to join/run/do lookups against the service, nns feels like an appropriate executable name for it, if it's not too much trouble to avoid a conflict for now? I've made some progress on the design, writeup soon. This week, though... what even is this week. upside_down_face
> 
> (haven't had a chance to review your code here again, just playing catch-up, but I could imagine this actually being the basis for the nns service in any event! I have no precious notions of us writing code from scratch vs contributing to your start here)

Renaming this currently internal tool to avoid conflicts is no problem -- pushing changes to rename this to `orb-ns` for lack of bikeshedding and using `orb` from the primary noosphere CLI.

@blaine the latest NS docs are [here](https://github.com/subconsciousnetwork/noosphere/blob/main/design/name-system.md), would appreciate any insight and feedback for sure, and would love to scope out your writeup on the protocol NNS when it's ready! The internal implementation of our NS ([noosphere-ns](https://github.com/subconsciousnetwork/noosphere/tree/main/rust/noosphere-ns)) has an agnostic DHT layer ([noosphere_ns::dht](https://docs.rs/noosphere-ns/latest/noosphere_ns/dht/index.html)), used by this little bootstrapper utility as well, that could be expanded to support something like "Protocol NNS" in either a generic or "standardized" sense (e.g. some UCAN/key patterns) -- happy to jam on some common ground and reusability here!

cdata (approved):

PR #161, opened 2022-11-16
Convert NameSystem/DHTNode to be generic over keys (DHTKeyMaterial).

jsantell:


cdata:
This looks good. I think this has marginal overlap with #162 . Would you be willing to try your hand at rebasing onto that change to see what the delta is? If it's not major, I might suggest we try to land #162 first (but if it is too hairy, we can land this one first).

jsantell:
@cdata No rebase issues here! Conflicts were in the orb-ns binary (rebased in #159)

cdata (approved):
Nice, the extra `Clone` bound is needed!

cdata (approved):

cdata (approved):
Let's gooo!

Issue #99, opened 2022-10-20
[p2p] Store DHT state in SphereFS


jsantell:
We should store DHT state, like recent peers, in SphereFS, rather than the out-of-the-box MemoryStore

cdata:
Linking to original feedback for additional context: https://github.com/subconsciousnetwork/noosphere/pull/93#discussion_r999934235

PR #169, opened 2022-11-19
dev: Introduce interior mutability for the NameSystem client

jsantell:
dev: Introduce interior mutability for the NameSystem client for querying from multiple threads, and introduce noosphere_core::data::Did in the NameSystem where appropriate, with additional PartialEq implementations for string comparison.

Some precursor work during the CLI integration of the NameSystem.

### DIFF
--- a/rust/noosphere-ns/src/name_system.rs
+++ b/rust/noosphere-ns/src/name_system.rs
@@ -6,9 +6,10 @@ use crate::{
 use anyhow::{anyhow, Result};
 use futures::future::try_join_all;
 use libp2p::Multiaddr;
-use noosphere_core::authority::SUPPORTED_KEYS;
+use noosphere_core::{authority::SUPPORTED_KEYS, data::Did};
 use noosphere_storage::{db::SphereDb, interface::Store};
 use std::collections::HashMap;
+use tokio::sync::{Mutex, MutexGuard};
 use ucan::crypto::did::DidParser;
 
 #[cfg(doc)]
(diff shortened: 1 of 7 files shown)


rust/noosphere-ns/src/name_system.rs
```diff
@@ -49,11 +50,11 @@ where
     pub(crate) key_material: K,
     pub(crate) store: SphereDb<S>,
     /// Map of sphere DIDs to [NSRecord] hosted/propagated by this name system.
-    hosted_records: HashMap<String, NSRecord>,
+    hosted_records: Mutex<HashMap<Did, NSRecord>>,
```
jsantell, line 53: This map isn't fully used (as auto-propagation isn't persistent), need to determine if persistence in libp2p is sufficient when backing via some store, where we do not need to manage "hosted records" in the client. #99

rust/noosphere-ns/src/name_system.rs
```diff
@@ -49,11 +50,11 @@ where
...
     /// Map of resolved sphere DIDs to resolved [NSRecord].
-    resolved_records: HashMap<String, NSRecord>,
+    resolved_records: Mutex<HashMap<Did, NSRecord>>,
     /// Cached DidParser.
-    did_parser: DidParser,
+    did_parser: Mutex<DidParser>,
```
jsantell, line 57: `DidParser` requires mutability; can remove once DidParser is reconfigured

Issue #167, opened 2022-11-18
`noosphere` tests are flakey


cdata:
Build where it failed mysteriously: https://github.com/subconsciousnetwork/noosphere/actions/runs/3498576071/jobs/5859116116

Summary of failure:

```
---- sphere::builder::tests::it_can_create_a_sphere_and_later_open_it stdout ----
thread 'sphere::builder::tests::it_can_create_a_sphere_and_later_open_it' panicked at 'called `Result::unwrap()` on an `Err` value: IO error: could not acquire lock on "/tmp/t2ec9-4/.sphere/storage/db": Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }', rust/noosphere/src/sphere/builder.rs:371:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

cdata:
Looks like maybe all of `noosphere` tests are haunted: https://github.com/subconsciousnetwork/noosphere/actions/runs/3498686886/jobs/5859368904

This very likely traces back to #162 :(

cdata:
These flakes are few and far between, but I'm sad to say we are still seeing them: https://github.com/subconsciousnetwork/noosphere/actions/runs/3567939081/jobs/5996231905

cdata:
Another flake: https://github.com/subconsciousnetwork/noosphere/actions/runs/3662928873/jobs/6192360285

cdata:
The same flake strikes again, but this time in `noosphere-cli`: https://github.com/subconsciousnetwork/noosphere/actions/runs/4077987629/jobs/7027661126

I wonder if `sled` is racing itself... maybe we can make sure it flushes before our storage is dropped?

jsantell:
Closing in favor of tracking IO lock errors in #579 (as none of the logs here are still around)

Issue #166, opened 2022-11-18
Investigate avoiding manual drop calls when using TempDir


jsantell:
In #162 and the NS bootstrap, we use the TempDir crate, with the former manually calling `drop()` to cleanup temp dirs, and the latter not doing so results in dangling dirs in `/tmp/`. Investigate to see why the TempDir requires manual dropping

cdata:
I think we have determined that we don't need to manually drop, and have stopped doing it as of #171

PR #171, opened 2022-11-21
chore: Use `tempfile` instead of `temp-dir`

cdata:
This is an attempt to clean up the way we create an dispose of temporary directories in the hopes it will help with #167 . Our temporary directory crate was deprecated, it turns out, so we update it here to [`tempfile`](https://crates.io/crates/tempfile). And, we fix a leak that caused temporary directories to go undeleted.

Even if this doesn't fix the flakes in `noosphere`, this is a hygienic improvement :shrug:

### DIFF
--- a/rust/noosphere/tests/integration.rs
+++ b/rust/noosphere/tests/integration.rs
@@ -47,7 +47,7 @@ async fn single_player_single_device_end_to_end_workflow() {
     #[cfg(target_arch = "wasm32")]
     tracing_wasm::set_as_global_default();
 
-    let (configuration, temporary_directories) = platform_configuration();
+    let (configuration, _temporary_directories) = platform_configuration();
     let key_name = "foobar";
 
     // Create the sphere and write a file to it
@@ -108,6 +108,4 @@ async fn single_player_single_device_end_to_end_workflow() {
 
         fs.save(None).await.unwrap();
     };
-
-    drop(temporary_directories)
 }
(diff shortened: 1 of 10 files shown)


rust/noosphere/tests/integration.rs
```diff
@@ -99,4 +108,6 @@ async fn single_player_single_device_end_to_end_workflow() {
 
         fs.save(None).await.unwrap();
     };
+
+    drop(temporary_directories)
```
jsantell, line 112: Is this needed? My understanding here (`drop` being an [empty function](https://doc.rust-lang.org/core/mem/fn.drop.html) that simply takes ownership) is this is redundant as `temporary_directories` goes out of scope here anyway

jsantell (approved):
LGTM! Should confirm that the `drop()` is necessary (#166), and note that `noosphere-ns`'s bin target uses `temp-dir` as well, which could use an upgrade too

rust/noosphere/tests/integration.rs
cdata: Yah, IDK what was happening before but I swear it was deleting the directories if I didn't manually drop... took the drops out and it's fine.

jsantell (dismissed):
LGTM!

cdata:
Had to rebase to fix conflicts

jsantell (approved):

Issue #176, opened 2022-11-27
Make IPFS syndication resilient to missing blocks


cdata:
Currently, we aren't able to recover gracefully when we try to sync revisions of content to IPFS with missing blocks. However, there are hypothetical cases where blocks may not be present in local storage but are possibly available elsewhere via IPFS.

For example, if I assign a slug in my local sphere to some content from another sphere (unlikely to happen from Subconscious, but still possible to do with Noosphere), we may encounter a case where the blocks of the foreign content are not yet available locally.

So, when syndicating to IPFS, we should try to fail gracefully in cases where blocks are missing from local storage.

Issue #175, opened 2022-11-27
Avoid costly, redundant IPFS syndication edge cases


cdata:
There will be cases where we probably don't want to syndicate content from a sphere to IPFS. For example, if a name directly references some other sphere, syndicating it would be costly and redundant. We should make sure that IPFS syndication is careful not to perform such redundant work.

Issue #156, opened 2022-11-16
Gateway API to publish a sphere version


cdata:
In our name system model, it is critical that a publishing event originate on a device with a key that has been authorized by the sphere to publish. So, in order for a name in our name system to be updated, a device with such a key must first generate a name record (a UCAN) to be published to the name system and then pass it off to a gateway. The gateway is then responsible for forwarding the name record to a name system.

We should expose a REST API in the gateway to support this flow.

This is necessary to implement #77

Issue #13, opened 2022-08-02
Shareable private data


cdata:
Our plan at this time is to implement a private data feature in Noosphere that enables users to author content that is private and/or only readable by an explicitly allowed audience.

This is the tracking issue for our progress towards shareable private data.

## User stories

As a Subconscious user, I want to write private notes and share them with specific collaborators, so I can collaborate with those I trust.

As a Subconscious user, I want to write notes that are shared only with mutuals, so I can share with my cozyweb friends, and discourage my notes going viral outside of that local context.

## Design notes

We intend to lean on end-to-end encryption strategies to enable this feature. In the default case, we assume that all data is publicly readable, but that some of that data is only made available in an encrypted state. Such encrypted data can be decrypted by the author, but may also be decrypted by an audience of others who the author explicitly addresses it to.

[Fission](https://fission.codes/) and [Peergos](https://peergos.org/) have done groundbreaking work building practical systems that achieve this quality, based upon the foundational concept of a [cryptree](https://ieeexplore.ieee.org/abstract/document/4032481). Fission's specialization of this is referred to as a [cryptDAG](https://whitepaper.fission.codes/file-system/partitions/private-directories/concepts/cryptree) and is implemented in their open source [Web Native File System](https://github.com/wnfs-wg/rs-wnfs/).

Serendipitously, Web Native File System is currently being re-written in Rust. This suggests that the shortest path to implementing shareable private data will be to incorporate WNFS into our design as it becomes practical to do so. WNFS is coherent with Noosphere as it is based on the same foundational concepts of CID-based content addressing and PKI.

Noosphere public content is non-hierarchical in nature, which means some of the qualities of of WNFS are not relevant for us in that case. It is possible (but not certain) that hierarchy has a place in Noosphere when it comes to shareable private data. To the extent that this is true, WNFS will seem more appealing. If it turns out that we can organize private data in useful ways without arbitrarily deep hierarchy, though, we may be able to implement a simplified solution to the problem.

gretchenfrage:
I hope I'm not sidetracking this issue too much as this may be too immature an idea to integrate into a project like this, but an idea I've liked for a while is to implement some sort of [HiStar](https://www.scs.stanford.edu/histar/)-inspired system to help prevent private data leakage from peer-to-peer node implementations.

[...]

PR #177, opened 2022-11-27
feat: Syndicate sphere revisions to IPFS Kubo

### LINKED ISSUE #163: Gateway should syndicate pushed blocks to IPFS
We are rounding the corner on integrating the Noosphere Name System with our gateway via #109, but in order for that to be useful for anything we need to make sure that sphere blocks are made available on IPFS.

To that end, we should implement support for a URL configuration for the gateway that tells the gateway where to send newly pushed blocks for pinning.

### LINKED ISSUE #172: `orb` should show a local sphere's identity somehow
In the original implementation of `orb`, we tracked a local sphere's identity in a file within the `.sphere` directory: `.sphere/IDENTITY` contained the DID.

As of #162 , we no longer render this value to a file (instead, it is kept as metadata in the local database). This makes it extremely hard for a user to get to their sphere's identity after it is first created. Even if that weren't the case, reading a file inside of `.sphere` is not very ergonomic for users in the first place. So, we should introduce a way for a user to learn the identity of their local sphere without manually diving through the `.sphere` directory.

cdata:
This change is our first step towards syndication of sphere content to IPFS. It implements a task that runs as a background thread in the gateway. The task waits for requests to syndicate a sphere version, and then syndicates all versions of the sphere that have not been syndicated _up to_ the given version. When a version is successfully syndicated, a checkpoint is updated -- any future jobs can start syndication from the last successful checkpoint.

This change implements syndication on every gateway push, but actually syndication should only occur on publish. We don't have a publish route yet, so I've marked this as work to be done when we implement publish in #156. I also filed #175 and #176 as future work that builds on this change.

As part of this change, I've introduced some interesting new dependencies to the `noosphere-cli` crate: one on https://github.com/wnfs-wg/rs-wnfs (for their Bloom filter implementation) and one on https://github.com/n0-computer/iroh (for their `iroh-car` package). Both of these dependencies are provisional; we may wish to explore alternatives for the library code we are using from them. But, we will attempt to use Rust WNFS as a storage backend for #13, so taking the dependency now helps us to test the waters for coherence across our libraries. In the case of Iroh, their CAR package comes close to what I wanted us to build for #2 , so I would like to explore that more in a future change. Note that the bulk of the lines changed here are from the new additions to `Cargo.lock` resulting from taking Git repository dependencies. The Bloom filter from rs-wnfs hasn't made it to a release yet, and after speaking with Iroh folks, it sounds like their packages will finally be published to crates.io next week. So in both cases, we should be able to remove the git dependency shortly.

In addition to the changes mentioned above, I've added an integration test for an end-to-end sync flow between two client spheres and a gateway. I realized that after #162 the CLI sync flow had broken without our tests failing. The new integration test should cover most of the broken cases. I also snuck in a fix for #172 - you can now get the local sphere's identity with `orb status`.

Fixes #172 
Fixes #163

### DIFF
--- a/rust/noosphere-cli/src/native/commands/serve/route/push.rs
+++ b/rust/noosphere-cli/src/native/commands/serve/route/push.rs
@@ -11,19 +11,22 @@ use noosphere_core::authority::{Authorization, SphereAction, SphereReference};
 use noosphere_core::data::Bundle;
 use noosphere_core::view::{Sphere, SphereMutation, Timeline};
 use noosphere_storage::{db::SphereDb, native::NativeStore};
+use tokio::sync::mpsc::UnboundedSender;
 use tokio::sync::Mutex;
 use ucan::capability::{Capability, Resource, With};
 use ucan::crypto::KeyMaterial;
 
-use crate::native::commands::serve::gateway::GatewayScope;
-use crate::native::commands::serve::{authority::GatewayAuthority, extractor::Cbor};
+use crate::native::commands::serve::{
+    authority::GatewayAuthority, extractor::Cbor, gateway::GatewayScope, ipfs::SyndicationJob,
+};
 
 // #[debug_handler]
 pub async fn push_route<K>(
     authority: GatewayAuthority<K>,
     ContentLengthLimit(Cbor(push_body)): ContentLengthLimit<Cbor<PushBody>, { 1024 * 5000 }>,
-    Extension(sphere_context): Extension<Arc<Mutex<SphereContext<K, NativeStore>>>>,
+    Extension(sphere_context_mutex): Extension<Arc<Mutex<SphereContext<K, NativeStore>>>>,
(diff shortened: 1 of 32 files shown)


rust/noosphere-cli/src/native/commands/serve/route/push.rs
```diff
@@ -117,6 +120,16 @@ where
         StatusCode::INTERNAL_SERVER_ERROR
     })?;
 
+    // TODO(#156): This should not be happening on every push, but rather on
+    // an explicit publish action. Move this to the publish handler when we
+    // have added it to the gateway.
+    if let Err(error) = syndication_tx.send(SyndicationJob {
```
jsantell, line 126: :+1:

jsantell (approved):
Great work. Interested in looking further at the syndication job work, great updates to docs as well here

PR #178, opened 2022-11-28
feat: Refactor storage interfaces

### LINKED ISSUE #174: Improve storage abstraction to allow for specialized blockstore
Hi,
here's a proposal that we briefly discussed on Discord today.

Currently, most of the codebase is generic over `S: Store`. The `Store` trait abstracts over key-value stores that take `key: &[u8], value: &[u8]`.  This makes it impossible to use an external block store with the frequently used `SphereDB` abstraction, because it is only generic over the `Store` and uses it for all stores, including the block store.

Instead, I propose to make the `SphereDB` and other parts of the code base generic over a `Storage` trait with associated types for (at least) a `KvStore` and a `BlockStore`. This allows implementers to use different storage backends for these stores.

In the future, more stores like the link store or the metadata could be specialized through further associated types on `Storage`.

The trait could look like this:

```rust
pub trait Storage {
    type BlockStore: BlockStore;
    type KvStore: KvStore;
    async fn get_kv_store(&self, name: &str) -> Result<Self::KvStore>;
    async fn get_block_store(&self, name: &str) -> Result<Self::BlockStore>;
}
```
As the `StorageProvider`  trait would then be used a lot to constrain generics, maybe it makes sense to rename it to just `Storage`.

The `SphereDb` could then be defined as follows:
```rust
pub struct SphereDb<S>
where
    S: Storage,
{
    block_store: S::BlockStore,
    link_store: S::KvStore,
    version_store: S::KvStore,
    metadata_store: S::KvStore,
}

````
The existing `Store` trait would be renamed to `KvStore`.

cdata:
The `noosphere-storage` crate is cleaned up and significantly refactored to support more ergonomic adaptation to other backend storage scenarios.

Special thanks to @Frando for suggesting this improvement. It was timely, as this same adaptability will be needed in upcoming web-based Noosphere experiments.

BREAKING CHANGE: The `StorageProvider` trait has been replaced by the `Storage` trait. This new trait allows for distinct backing implementations of `BlockStore` and `KeyValueStore`.

BREAKING CHANGE: A `SphereDb` was previously generic over a `Store`, but is now generic over the new `Storage` trait.

BREAKING CHANGE: The modules in the `noosphere-storage` crate have been re-arranged so that they are organized more consistently with the way we organize modules in other crates.

Fixes #174


jsantell (dismissed):
Great changes! Nice documentation updates. Only thing I have to suggest, and not sure worth the change, but consider uplifting storage objects to crate-level, as they're used often throughout the system e.g. `noosphere_storage::{MemoryStorage, SphereDb, KeyValueStorage}` rather than copying/remembering 3-4 individual paths in the tree `noosphere_storage::{memory::MemoryStorage, db::SphereDb, key_value::KeyValueStorage}`

cdata:
I have to rebase this change anyway, so I'll apply your suggestion @jsantell

jsantell (approved):
New crate level paths are :kissing_closed_eyes: :ok_hand: 

Looks great!

PR #180, opened 2022-11-29
chore: Update IPLD-adjacent dependencies

cdata:
BREAKING CHANGE: Several critical dependencies of this library were updated to new versions that contain breaking changes.

### DIFF
--- a/rust/noosphere-ns/src/records.rs
+++ b/rust/noosphere-ns/src/records.rs
@@ -202,7 +202,7 @@ impl FromStr for NSRecord {
         // Wait for next release of `ucan` which includes traits and
         // removes `try_from_token_string`:
         // https://github.com/ucan-wg/rs-ucan/commit/75e9afdb9da60c3d5d8c65b6704e412f0ef8189b
-        Ok(NSRecord::new(Ucan::try_from_token_string(ucan_token)?))
+        Ok(NSRecord::new(Ucan::from_str(ucan_token)?))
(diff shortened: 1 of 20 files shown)


rust/noosphere-ns/src/records.rs
```diff
@@ -202,15 +202,15 @@ impl FromStr for NSRecord {
         // Wait for next release of `ucan` which includes traits and
         // removes `try_from_token_string`:
         // https://github.com/ucan-wg/rs-ucan/commit/75e9afdb9da60c3d5d8c65b6704e412f0ef8189b
-        Ok(NSRecord::new(Ucan::try_from_token_string(ucan_token)?))
```
jsantell, line 205: Can remove this comment now with the new UCAN

jsantell (approved):
LGTM! There's an extra comment/todo about the UCAN update in `NSRecord` that might be worth taking a look

rust/noosphere-ns/src/records.rs
jsantell: Might be some other changes via the addition of traits in ucan package, I can take a look separately

jsantell (approved):

rust/noosphere-ns/src/records.rs
cdata: If it's alright, I'll let you take care of this one for me :pray:

PR #182, opened 2022-12-03
feat: Introduce web bindings and `orb` NPM package

### LINKED ISSUE #165: Introduce an NPM package built around Noosphere Wasm library
As a follow-up to #164 and in the spirit of #131 , we should introduce an NPM package (that is, a JavaScript package built from TypeScript sources) that wraps the Noosphere Wasm library (and associated TypeScript bindings) in a high-level API that is suitable for general consumption by web developers who are comfortable wielding the NPM package ecosystem.

The package should probably load and run the Wasm core in a web worker, while the high-level API may be wielded from the main thread or another worker.

cdata:
This change proposes a milestone feature for Noosphere: a working NPM package that vends a TypeScript-based library that exposes Noosphere to JavaScript via a Wasm kernel.

In addition to the NPM package, this change includes `wasm-bindgen`-based FFI on the Rust side and a minimum viable Lit + Redux viewer application that is designed to render sphere content when loaded from a commodity IPFS Kubo gateway.

In support of the loading-out-of-a-gateway use case, the change includes a feature-gated `KuboStorage` implementation that can wrap any `Storage` implementation. `KuboStorage` will "fall back" to looking for blocks from an IPFS gateway if a requested block is not found in local storage. This makes `KuboStorage` suitable for read-only Noosphere use cases where an IPFS gateway API may be available.

I chose https://github.com/google/wireit to orchestrate builds across Rust and TypeScript packages. Largely thanks to Wireit, we don't need to be aware of yet-another-tool to run correct builds that span multiple source trees. Running `npm run build` from the `typescript` folder is sufficient to build everything (including Rust dependencies) in the intended order. The same goes for running tests: `npm run test` from the `typescript` folder will ensure all dependencies are built in time for the tests to run.

@aomarks has offered to lend us his scrutiny to make sure we get the most out of `wireit`.

Our test suite has also been expanded to include TypeScript tests (which are Mocha-formatted tests that run in headless Chrome). This should get us coverage from the JavaScript side of the web FFI.

Fixes #165


typescript/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "description": "",
+  "workspaces": ["./packages/*"],
+  "scripts": {
+    "build": "wireit",
+    "clean": "wireit",
+    "serve": "wireit",
+    "test": "wireit"
+  },
+  "dependencies": {
+    "wireit": "^0.8.0",
```
aomarks, line 16: Can upgrade to `^0.9.0`

typescript/packages/orb/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "build": "wireit",
+    "generate_wasm": "wireit",
+    "clean": "wireit",
+    "test": "wireit"
+  },
+  "dependencies": {
+    "@web/dev-server": "^0.1.34",
+    "esbuild": "^0.15.9",
+    "typescript": "~4.8.2",
+    "wireit": "^0.8.0"
```
aomarks, line 19: Can upgrade to `^0.9.0`

typescript/packages/orb/package.json
```diff
@@ -0,0 +1,58 @@
...
+  },
+  "wireit": {
+    "build": {
+      "dependencies": [
+        "generate_wasm"
+      ],
+      "command": "tsc",
+      "files": [
+        "./src/**/*.ts"
+      ]
```
aomarks, line 29: Missing `output` here? If you add it, you'll get incremental build and caching (e.g. if none of the input files changed since the last run, it won't re-run).

typescript/packages/orb/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "esbuild": "^0.15.9",
+    "typescript": "~4.8.2",
+    "wireit": "^0.8.0"
+  },
+  "wireit": {
+    "build": {
+      "dependencies": [
+        "generate_wasm"
+      ],
+      "command": "tsc",
```
aomarks, line 26:
```suggestion
      "command": "tsc --pretty",
```

Since `tsc` won't think it's attached to a TTY when launched via wireit, it will turn off its nice colorful output by default.

typescript/packages/orb/package.json
```diff
@@ -0,0 +1,58 @@
...
+      "files": [
+        "./src/**/*.ts"
```
aomarks, line 28: Add `tsconfig.json` and `../../tsconfig.base.json` here so that if you change a setting, wireit will know a rebuild is needed.

typescript/packages/orb/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "test": "wireit"
+  },
+  "dependencies": {
+    "@web/dev-server": "^0.1.34",
+    "esbuild": "^0.15.9",
+    "typescript": "~4.8.2",
+    "wireit": "^0.8.0"
+  },
+  "wireit": {
+    "build": {
```
aomarks, line 22: See https://github.com/google/wireit#typescript for a bunch of suggestions for `tsc` that can improve performance.

typescript/packages/orb/package.json
aomarks, line 29: Adding `"clean": "if-file-deleted"` is nice for `tsc`, for the case where you remove a source file, because `tsc` doesn't cleanup stale output artifacts by itself.

typescript/packages/orb/package.json
```diff
@@ -0,0 +1,58 @@
...
+        "../../../rust/**/Cargo.toml",
+        "../../../rust/**/*.rs",
+        "./scripts/genergate-wasm-artifacts.sh"
+      ],
+      "output": [
+        "lib/noosphere*.{ts,wasm,js}"
+      ]
+    },
+    "clean": {
+      "command": "rm -rf ./lib && rm -f ./tsconfig.tsbuildinfo"
```
aomarks, line 44: I'd suggest leaning on the automatic cleaning features of wireit instead of doing this manually: https://github.com/google/wireit#cleaning-output

typescript/packages/orb/package.json
```diff
@@ -0,0 +1,58 @@
...
+    },
+    "test": {
+      "dependencies": [
+        "build"
+      ],
+      "command": "web-test-runner --node-resolve"
```
aomarks, line 50:
```suggestion
      "command": "web-test-runner --node-resolve",
      "files": ["web-test-runner.config.js"],
      "output": []
```

This way if no input files changed, a previously passing result will be remembered, and the test won't need to run again until you change something.

The `"output": []` is required because wireit treats "unknown output files" differently from "definitely no output files".

typescript/packages/sphere-viewer/package.json
```diff
@@ -0,0 +1,49 @@
...
+      ],
+      "command": "tsc"
+    },
+    "serve": {
+      "dependencies": [
+        "build"
+      ],
+      "command": "web-dev-server -r ./dist --open"
+    },
+    "clean": {
```
aomarks, line 45: See other comment above about using built-in cleaning feature.

typescript/packages/sphere-viewer/package.json
```diff
@@ -0,0 +1,49 @@
...
+      "dependencies": [
+        "../orb:build"
+      ],
+      "command": "tsc"
+    },
+    "serve": {
+      "dependencies": [
+        "build"
+      ],
+      "command": "web-dev-server -r ./dist --open"
```
aomarks, line 43: You likely want `"service": true` here, so that you can run `npm run serve --watch`. Without `"service": true`, watch mode won't work because Wireit will be waiting for the `web-dev-server` process to exit, but it never will.

https://github.com/google/wireit#services

typescript/packages/sphere-viewer/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "typescript": "~4.8.2",
+    "vite": "^3.2.3",
+    "vite-plugin-wasm": "^3.1.0",
+    "wireit": "^0.8.0"
+  },
+  "wireit": {
+    "build": {
+      "dependencies": [
+        "build:tsc"
+      ],
```
aomarks, line 30:
```suggestion
      "dependencies": [
        {
          "script": "build:tsc",
          "cascade": false
        }
      ],
```

You may want to disable "cascade" here, because most likely this server doesn't need to restart if you're in watch mode and these assets changed (but you still do want the dependency because you want them to be kept up to date).

See item 2 at https://github.com/google/wireit#reasons-to-disable-cascade

typescript/packages/sphere-viewer/package.json
```diff
@@ -0,0 +1,49 @@
...
+      "command": "vite build"
+    },
+    "build:tsc": {
+      "dependencies": [
+        "../orb:build"
+      ],
+      "command": "tsc"
```
aomarks, line 37: See other comments about `tsc` above, and https://github.com/google/wireit#typescript for some general advice on `tsc`.

.github/workflows/run_test_suite.yaml
```diff
@@ -99,3 +99,51 @@ jobs:
...
+          curl -L -O "https://chromedriver.storage.googleapis.com/${CHROMEDRIVER_VERSION}/chromedriver_linux64.zip"
+          unzip chromedriver_linux64.zip && chmod +x chromedriver && sudo mv chromedriver /usr/local/bin
+          chromedriver -version
+        shell: bash
+      - name: 'Install NPM dependencies'
+        working-directory: ./typescript
+        run: npm ci
+      - name: 'Run TypeScript headless browser tests'
+        working-directory: ./typescript
+        run: npm run test
```
aomarks, line 149: Could consider enabling automatic GitHub Actions caching. This would restore your output from cache across CI iterations when the input files haven't changed, and if nothing in the transitive dependency graph for the test has changed, the test won't need to run at all.

https://github.com/google/wireit#github-actions-caching

aomarks (commented):
🎉🎉🎉

Awesome!

Many of these suggestions are optional, in particular the ones around including `files` and `output`, because it is valid to not include those, but without them you won't take full advantage of incremental build/caching.

typescript/packages/orb/package.json
cdata: It's a bit tedious to anticipate all of the outputs in cases like this. Are there any shortcuts to, say, derive outputs automatically from the list of inputs? Like: every input .ts file produces a corresponding output .d.ts and .js file.

rust/noosphere/src/sphere/builder.rs
```diff
@@ -106,7 +110,14 @@ impl SphereContextBuilder {
     /// Specify the URL of a gateway API for this application to sync sphere
     /// data with
     pub fn syncing_to(mut self, gateway_url: Option<&Url>) -> Self {
-        self.gateway_url = gateway_url.cloned();
+        self.gateway_api = gateway_url.cloned();
+        self
+    }
+
+    /// Speciy the URL of an IPFS Kubo RPC API for this application to access
```
jsantell, line 117: nit: Specify

rust/noosphere/src/wasm/fs.rs
```diff
@@ -0,0 +1,143 @@
...
+            .map_err(|error| format!("{:?}", error))?;
+
+        Ok(file.map(|file| SphereFile {
+            inner: file.boxed(),
+        }))
+    }
+
+    #[wasm_bindgen(js_name = "writeString")]
+    /// Write content to a name in the sphere's namespace as a UTF-8 encoded
+    /// string
```
jsantell, line 39: Does wasm bindgen handle WTF-16 => UTF-8 conversions?

rust/noosphere/src/wasm/sphere.rs
```diff
@@ -0,0 +1,46 @@
...
+
+#[wasm_bindgen]
+/// A `SphereContext` is a view into all of a sphere's data, that also
+/// encapsulates handles to local storage and a user's authority relative to the
+/// sphere. If a user is appropriately authorized, they may use a
+/// `SphereContext` to modify a sphere. Otherwise, they may only read a sphere's
+/// publicly visible content.
+pub struct SphereContext {
+    #[wasm_bindgen(skip)]
+    pub inner: Arc<Mutex<SphereContextImpl<PlatformKeyMaterial, PlatformStorage>>>,
```
jsantell, line 21: How do mutexes work via wasm bindgen? Atomics and SABs? Curious about the runtime implementation of this!

typescript/packages/orb/package.json
aomarks: Having a dedicated output folder could be a simple solution, since then you can just put e.g. `lib`

I have been considering a special mode where it runs each script in isolation, looks for new files, and then suggests updates to your config (and/or warns you if there are undocumented outputs).

jsantell (commented):
Just looking through the Rust parts, it's looking excellent

rust/noosphere-cli/src/bin/orb.rs
```diff
@@ -8,6 +8,7 @@ pub async fn main() -> anyhow::Result<()> {
 }
 
 #[cfg(target_arch = "wasm32")]
+#[allow(unused_must_use)]
```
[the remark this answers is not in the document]
cdata: Because this noop always gives us a warning when this crate is built (which it always is) for the `wasm32-unknown-unknown` target.

rust/noosphere/src/wasm/fs.rs
cdata: I believe so, yes. JS `string` to Rust `String` conversion is handled in the codegen.

rust/noosphere/src/wasm/sphere.rs
cdata: Yeah, `wasm-bindgen` brings along a single-threaded async runtime. The fact that we can drop-in some alternative runtime is the silver lining of super-low-level Rust async, and so we take advantage of it here. It generally means that anything we might have otherwise been able to parrallelize using futures becomes at best concurrent.

typescript/packages/orb/package.json
cdata: The challenge in this particular case is that some of what ends up in `lib` is produced by `wasm-bindgen`. For now, I'm manually listing all the outputs.

typescript/packages/orb/src/index.test.ts
```diff
@@ -0,0 +1,134 @@
...
+describe('orb', () => {
+  it('can initialize a NoosphereContext', () => {
+    const noosphere = new NoosphereContext('foo');
+
+    noosphere.free();
+  });
+
+  it('can create a key', async () => {
+    const noosphere = new NoosphereContext('foo');
+    const keyName = makeRandomName('key');
```
jsantell, line 26: I think it'd be valuable confirming that non-ascii/UTF8 JS strings work as intended over wasm (as opposed to a TextDecoder + byte writing method)

rust/noosphere/src/wasm/sphere.rs
jsantell: That sounds way easier than COOP/COEP configuration, great

typescript/packages/orb/src/index.test.ts
cdata: SG will verify

cdata: Good thing you asked about this, there are some considerations to be made. See: https://rustwasm.github.io/docs/wasm-bindgen/reference/types/str.html#utf-16-vs-utf-8

TL;DR the conversion is mostly lossless, but lossy around the edges.

I think in the end we probably would want to stick with the two methods as written, since they give our users two options with different trade-offs. If a user wants to control the conversion, they can encode the bytes themselves on the JS side and pass them in. If they are okay with the conversion as performed by our codegen, then they can pass a string.

I'll add comments to this change to advise of the caveats.

typescript/packages/sphere-viewer/src/index.ts
```diff
@@ -0,0 +1,58 @@
...
+    })
+    .reduce((map: { [index: string]: string }, entry) => {
+      if (entry[0] && entry[1]) {
+        map[entry[0]] = entry[1];
+      }
+      return map;
+    }, {});
+}
+
+const queryState = getQueryState();
```
SebastianZimmer, line 22: Optional simplification idea:

```suggestion
const queryState = Object.fromEntries(new URLSearchParams(location.search));
```

jsantell (approved):

typescript/packages/orb/src/index.test.ts
jsantell: SGTM, I think documenting the string method with the rust wasm encoding note is sufficient

cdata:
I'm going to go ahead and merge since the last approval was dismissed by an unambiguously good (and small) patch. Thanks so much for the feedback @jsantell @aomarks and @SebastianZimmer !

PR #184, opened 2022-12-06
Make transclude font size same as body font size

gordonbrander:
Removes CSS that made transclude text smaller than body text. This looks better, in practice.


cdata (approved):
omg it is so beautiful :heart_eyes:

PR #191, opened 2022-12-13
dev: Integration of NameSystem within the CLI v1

jsantell:
* Added a NS jobs thread, similar to the IPFS syndication worker. Completed get_record queries are stored in gateway sphere.
* Added barebones PUBLISH route, which triggers an NS set_record job.
* Added a hook to NS get_record from the PUSH route.
* Added inner mutability to the core NameSystem object.
* Created an in-progress TaskRunner for resiliant jobs, not yet used by the IPFS/NS workers.
* The fetching of records on PUSH is not yet completed, though so prelim work here.


rust/noosphere-cli/tests/gateway.rs
```diff
@@ -33,11 +32,92 @@ use noosphere_cli::native::{
...
+        .build()?;
+    node.connect().await?;
+    let address = node.p2p_address().unwrap().to_owned();
+    Ok((key, db, node, vec![address]))
+}
+
+/// Copies the authorization UCAN associated with `context` into the `dest_store`.
+/// Sort of a hack to propagate the CID from local sphere to the name system's
+/// store in order to validate.
+async fn copy_authorization_ucan_token<K: DHTKeyMaterial + 'static, S1: Storage, S2: Storage>(
```
jsantell, line 65: Unsure of the best way to propagate the delegation CID to bootstrap NS. If PUBLISH called after a PUSH, the delegation should be available on IPFS, though the NS doesn't have a running KuboClient or anything yet. So still a few things need to be thought out w/r/t validating things in the NS relying on IPFS..

rust/noosphere-core/src/authority/author.rs
```diff
@@ -103,6 +107,34 @@ where
 
         Ok(Access::ReadOnly)
     }
+
+    /// A helper for creating and signing a new UCAN token for publishing.
+    pub async fn create_publish_token(
```
jsantell, line 112: This seems like a nice function to have along with the NS/DHT crate, though designed to be non-web, wonder if we can still expose some data types like `NSRecord` or the token gen utilities, or maybe these are core capabilities.

rust/noosphere-cli/src/native/commands/serve/name_system/tasks.rs
```diff
@@ -0,0 +1,357 @@
+///! This task manager is not currently used, but a potential
```
jsantell, line 1: Need to give this some more thought and jam on how we can use this, but unused at the moment.

jsantell:
Superceded by #253

PR #192, opened 2022-12-13
doc: Add mermaid graph of Rust module dependencies

jsantell:


rust/README.md
```diff
@@ -13,6 +13,30 @@ Most other language implementations of Noosphere bind to this Rust
...
+
+noosphere((Noosphere))
+cli((CLI))
+
+noosphere --> api
+noosphere --> core
+noosphere --> fs
+noosphere --> storage
+noosphere --> into
+cli --> api
```
cdata, line 29: I think this isn't quite right; the `cli` depends most critically on `noosphere`.

cdata (changes requested):

rust/README.md
jsantell: Nice catch!! Updated

cdata:
We may want to revise this, but this is a very cool addition to the README so let's go ahead and land it :raised_hands:

cdata (approved):

PR #193, opened 2022-12-14
fix: Recover from Kubo pin check

cdata:
Previously, if a Kubo instance was missing a pin, we would get an unexpected response and the client method would return an error result. Now we recover and just mark the block as not pinned (it is acceptable to resend a block that is actually pinned, just a wasteful).


jsantell (approved):
LGTM

PR #196, opened 2023-01-05
[dev] Improvements to the NameSystem based on initial gateway integration

jsantell:
* Added inner mutability to the core NameSystem object for use in multithreaded gateway.
* Decouple bootstrap peers and listener from NameSystem construction. A challenge when creating a network of peers, a peer's address is needed before connecting and peering, needing to resolve a default port of 0 to the real port, into a shareable, Multiaddr P2P address.
* Default listening port now leverages libp2p's port selection. With parallel tests running on live ports, random port selection caused intermittent port conflicts, hardcoding unique ports is tedious and fragile, and recreating libp2p's default port selection is redundant. Tests and scenarios where specific ports do not matter can simply listen to TCP port 0, and a default is selected.
* Migrate NS CLI to use NameSystem directly rather than the underlying DHT nodes for consistency as a single entry point to NSS, ensuring that the gateway, NSS CLI/bootstrap nodes, and tests all use the same high level API, configurations, and validations.
* Ownership/API cleanups. With a better understanding of the DHT implementation and the above changes, generics from NameSystem and DHTNode can be removed. Function names of NS and DHT more strongly aligned, and more explicit of ownership (e.g. prefer a non-reference over a reference that gets cloned). Better isolation of Storage objects between network nodes in the same process.

This absorbs changes from #169, and integration (#191) should be rebased on this.


rust/noosphere-ns/src/bin/orb-ns/runner/config.rs
```diff
@@ -65,22 +60,32 @@ impl RunnerConfig {
             CLICommand::Run {
                 config,
                 key,
+                mut bootstrap,
                 mut port,
             } => match config {
                 Some(config_path) => {
                     let toml_str = tokio::fs::read_to_string(&config_path).await?;
-                    let config: CLIConfig = toml::from_str(&toml_str)?;
+                    let config: CLIConfigFile = toml::from_str(&toml_str)?;
```
cdata, line 68: Will the NS be able to be initialized before a configuration file is present?

cdata (approved):
This looks like a great step forward.

rust/noosphere-ns/src/bin/orb-ns/runner/config.rs
jsantell: The CLI flags support defining a listening address and bootstrap peers, though limited to one node -- the config file has additional configurations possible, namely DHTConfig values (e.g. libp2p timeouts and knobs) and multiple nodes per parent process (unclear how useful multiple nodes are on a single machine, though more of a stopgap until nodes are e.g. across the globe). DHTConfig values can be mapped to CLI flags of course if it looks to be valuable. Is that the scenario you're thinking of? (the "before" is confusing me a bit)

PR #202, opened 2023-01-19
feat: Extend C FFI for header enumeration

cdata:
This change introduces two new C FFI functions for dealing with headers, and an example of how to use the API from Swift as a test.

### DIFF
--- a/rust/noosphere/src/ffi/fs.rs
+++ b/rust/noosphere/src/ffi/fs.rs
@@ -1,4 +1,5 @@
 use anyhow::anyhow;
+use itertools::Itertools;
 use noosphere_core::data::Did;
 use noosphere_fs::{SphereFile, SphereFs};
 use safer_ffi::prelude::*;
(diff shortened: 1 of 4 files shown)


rust/noosphere/src/ffi/fs.rs
```diff
@@ -241,3 +242,36 @@ pub fn ns_sphere_file_header_values_read(
...
+        .memo
+        .get_first_header(name.to_str())
+        .into_iter()
+        .nth(0)
+        .map(|value| value.try_into().unwrap())
+}
+
+#[ffi_export]
+/// Read all the headers associated with a file as an array of strings. Note
+/// that headers will be reduced to a single entry in cases where multiple
```
gordonbrander, line 263: If I understand right, we drop duplicate headers, rather than taking the HTTP approach of concatenating them with a comma?

https://www.rfc-editor.org/rfc/rfc2616.html#section-4.2

> Multiple message-header fields with the same field-name MAY be
   present in a message if and only if the entire field-value for that
   header field is defined as a comma-separated list [i.e., #(values)].
   It MUST be possible to combine the multiple header fields into one
   "field-name: field-value" pair, without changing the semantics of the
   message, by appending each subsequent field-value to the first, each
   separated by a comma. The order in which header fields with the same
   field-name are received is therefore significant to the
   interpretation of the combined field value, and thus a proxy MUST NOT
   change the order of these field values when a message is forwarded.

FWIW, I think dropping duplicate headers is probably preferable for now. I feel we want to discourage the comma-separated value headers, since it makes assumptions about the header value format.

gordonbrander: Ah, nvm this only retrieves unique header KEYS. That's perfect.

rust/noosphere/src/ffi/fs.rs
```diff
@@ -241,3 +242,36 @@ pub fn ns_sphere_file_header_values_read(
         .into_boxed_slice()
         .into()
 }
+
+#[ffi_export]
+/// Get the first header value for a file that corresponds to the given name, if any
```
jsantell, line 247: nit: *Get the first header value that corresponds to the given name for a file.*

swift/Tests/SwiftNoosphereTests/NoosphereTests.swift
```diff
@@ -58,4 +58,68 @@ final class NoosphereTests: XCTestCase {
...
+        // NOTE: "hello" is only given once even though there are two headers with that name
+        assert(name_count == 3)
+        
+        for i in 0..<name_count {
+            let name = String.init(cString: pointer.pointee!)
+            let first_value_ptr = ns_sphere_file_header_value_first(file, name)
+            let first_value = String.init(cString: first_value_ptr!)
+            ns_string_free(first_value_ptr)
+            
+            print("Header name:", name)
```
jsantell, line 108: `first_value` is expected to be "bar", "world", "world" here, consider asserting that

rust/noosphere/src/ffi/fs.rs
jsantell: No comment on dropping dupes vs multiple headers, but the current Swift test has a header entry `"hello:noosphere"` that cannot be retrieved with the current FFI as an example; whether dupe headers or dropping dupes, the structs should reflect that

cdata: It can be retrieved with the API that existed before these helpers were added @jsantell `ns_sphere_file_header_values_read` gives you all values for a name.

cdata:
Gonna go ahead and merge this with small tweaks in place from feedback. It's a small change and needs to be released in a timely fashion.

rust/noosphere/src/ffi/fs.rs
jsantell: perfect!!

PR #207, opened 2023-01-25
[feat] DHT configuration and status API

jsantell:
* Adds `server::APIServer`, a web server for local interaction of name system node.
* Adds `server::HTTPClient` for interfacing with `server::APIServer`.
* `NameSystemClient` trait introduced, implemented by both `NameSystem`
  `and server::HTTPClient`.
* Uses my own fork of libp2p (0.51.0) with upstream changes in the interim.

Remaining areas of additional work for orb-ns CLI:
  * Test get/put records with IPFS
  * Introduce noosphere-ipfs to DHT processor. Test.
  * Expose HTTPClient to local node via orb-ns.


rust/noosphere-ns/src/bin/orb-ns/runner/mod.rs
```diff
@@ -1,5 +1,108 @@
...
+        )?;
+
+        // Request address from DHT to resolve default port (0) to
+        // selected port.
+        let address = node
+            .listen(node_config.listening_address.to_owned())
+            .await?;
+        node.add_peers(node_config.peers.to_owned()).await?;
+
+        println!("Listening on {}...", address);
```
cdata, line 63: Just a general note to prefer `tracing::*` logging over `println`

rust/noosphere-ns/tests/ns_test.rs
```diff
@@ -4,15 +4,16 @@ pub mod utils;
 use anyhow::Result;
 use noosphere_core::{authority::generate_ed25519_key, data::Did, view::SPHERE_LIFETIME};
 use noosphere_ns::{
-    utils::{generate_capability, generate_fact},
-    Multiaddr, NSRecord, NameSystem,
+    utils::{generate_capability, generate_fact, wait_for_peers},
+    Multiaddr, NSRecord, NameSystem, NameSystemClient,
```
cdata, line 8: Sorry that this occurred to me late, but technically the Rust naming conventions have this to say about acronyms:

> In UpperCamelCase, acronyms and contractions of compound words count as one word: use Uuid rather than UUID, Usize rather than USize or Stdin rather than StdIn. In snake_case, acronyms and contractions are lower-cased: is_xid_start.

https://rust-lang.github.io/api-guidelines/naming.html

So, for example, we write `IpfsClient` instead of `IPFSClient`. So, we should write `NsRecord` instead of `NSRecord`.

cdata (approved):
This looks great. I would be happy to see the suggested changes take place in follow-up work.

PR #209, opened 2023-01-25
dev: upgrade libp2p to 0.50.0

jsantell:
Bumping libp2p from 0.49.0 to 0.50.0:

* Update to latest APIs.
* Paving way to integrate uplifts https://github.com/libp2p/rust-libp2p/pull/3387
* Reconfigure the `NetworkBehaviour` derive macro. Previously, analyzer would alert warnings and fail (multiple machines) with this macro, but enabling the `macros` feature works like a charm (maybe this was optional before).
* Remove some workarounds in lieu of libp2p features (e.g. we don't have to roll our own Tokio executor anymore)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,11 +4,32 @@ version = 3
 
 [[package]]
 name = "aead"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


cdata (approved):
Looks great!

PR #216, opened 2023-02-02
fix: Ensure that sphere changes exclude `since`

cdata:
This was causing one too many "changes" to be reported by the corresponding sphere APIs.


jsantell (approved):

gordonbrander (approved):
Fix looks good to me.

jsantell (approved):

PR #219, opened 2023-02-07
feat: General error handling in C FFI

cdata:
This change is a follow-up to #210 where we introduced a new error handling style to the FFI. This change completes the work, changing the rest of our FFI to support the same error handling style.


jsantell (dismissed):
very nice

PR #221, opened 2023-02-08
chore: Minimum viable container builds

cdata:
This change is small but several weeks in the making. It implements a Github workflow for building and pushing containers to a private registry on GCP, from which we may deploy them as part of a managed service for the benefit of our users.

Containers are automatically built and pushed for each PR, and are available via the tag `pr-$PR_NUMBER`. The same workflow may be dispatched (manually or via a caller workflow) to build and push containers against arbitrary tag names. So, eventually we can integrate this with our release workflow.

### DIFF
--- a/.github/workflows/container_images.yaml
+++ b/.github/workflows/container_images.yaml
@@ -4,22 +4,62 @@ on:
     inputs:
       image-tag:
         type: string
-      # default: ${{ github.sha }}
   workflow_dispatch:
     inputs:
       image-tag:
         type: string
-        # default: ${{ github.sha }}
 
 name: 'Container images'
 
 jobs:
+  determine-image-tags:
+    name: 'Determine image tags'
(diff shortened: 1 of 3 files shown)


.github/workflows/container_images.yaml
```diff
@@ -0,0 +1,94 @@
...
+    name: 'Build and push'
+    runs-on: ubuntu-latest
+    needs: ['determine-image-tags']
+    concurrency: 'docker-build'
+    permissions:
+      contents: 'read'
+      id-token: 'write'
+    strategy:
+      matrix:
+        image: [{ name: 'orb', file: 'images/orb/Dockerfile' }]
```
cdata, line 61: Add an entry to this JSON-y value to include another container build in the workflow

jsantell (approved):
LGTM!

PR #222, opened 2023-02-09
dev: Follow up of initial orb-ns implementation.

jsantell:
* Add tests for CLICommands using APIServer and HTTPClient.
* Remove superfluous ctrl-c signal handling causing issues.
* Begin normalizing rendering of CLI output as JSON strings, or wrapped JSON error strings.
* Return runner information on API address in order to use dynamic ports in testing.

PR #223, opened 2023-02-10
dev: initial containerization work for orb-ns

jsantell:
Lots of different ways to handle parameterization here, but a good first pass


jsantell:
* Updated with non-root user
* Updated with `0.0.0.0`, though expecting all parameterization to change
* Needs rebase on #222

cdata (approved):
Great work! Already seeing the changes show up in the GCP registry :D

PR #231, opened 2023-02-16
feat: Always flush on SphereFS save

cdata:
This should help us rule out durability issues in our underlying K/V store that may be cropping up when the Subconscious app crashes.

### DIFF
--- a/rust/noosphere-storage/src/db.rs
+++ b/rust/noosphere-storage/src/db.rs
@@ -101,6 +101,31 @@ where
         self.version_store.get_key(identity).await
     }
 
+    /// Manually flush all pending writes to the underlying [Storage]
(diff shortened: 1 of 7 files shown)


rust/noosphere-storage/src/db.rs
```diff
@@ -101,6 +101,31 @@ where
...
+            ("link", link_store_result),
+            ("version", version_store_result),
+            ("metadata", metadata_store_result),
+        ];
+
+        for (store_kind, result) in results {
+            if let Err(error) = result {
+                warn!("Failed to flush {} store: {:?}", store_kind, error);
+            }
+        }
```
jsantell, line 124: Clean :sparkles:

jsantell (approved):
Speedy fix, LGTM

PR #234, opened 2023-02-17
dev: Support ephemeral keys in orb-ns docker image.

jsantell:


cdata (approved):
This looks good. At some point, we should probably review the CLI interface to line it up a little more closely with `orb` (e.g., `orb key create $name`).

jsantell:
> This looks good. At some point, we should probably review the CLI interface to line it up a little more closely with `orb` (e.g., `orb key create $name`).

Absolutely! An alignment pass is needed once everything is clicking together. Specifically, `orb-ns` does not *require* a key-gen method, but it reads keys the same way as `orb` does, so having the counterpart was useful during dev. Possible paths forward:

- Keep It Simple: Remove key functionality from `orb-ns`; accept an arg or file with private key (+1 for devops currently, though extracting keys from non InsecureKeyStorage e.g. via CLI may be challenging in future key storage scenarios)
- Quick & dirty: Reimplement some subset of `orb key` for `orb-ns`
- Build out CLI ecosystem: Package `orb key` as `orb-key`, make available under both `orb key` and `orb-ns key`
- Mono CLI: Package `orb-key` and `orb-ns` under `orb key` and `orb ns` respectively, git-style (`git status` -> `git-status`) -- As `ns` is a dependency of `orb`, this has some good plus sides: a single CLI binary for users and infra, reusing mechanisms between the two; the only scenario where this is worse AFAICT is increased binary size when running bootstrap nodes outside of a gateway, which isn't a use case currently, and a workaround there is simply a "gateway" feature flag. This is my lofty preference I think.

PR #238, opened 2023-02-20
feat: Introduce `noosphere-gateway` crate

cdata:
This change moves the gateway functionality of the CLI out of the `noosphere-cli` crate and into its own package. This has been a long time coming, since every other corner of the CLI has been a light wrapper over functionality in other crates since #162 .

This change is a precursor to integrating the new name system work into the gateway.


rust/noosphere-core/src/tracing.rs
```diff
@@ -0,0 +1,28 @@
+///! Helpers to wrangle logging across Noosphere crates
+///! NOTE: [initialize_tracing] should only ever be called in tests or binaries;
+///! a library should only concern itself with instrumentation and logging.
+use std::sync::Once;
```
jsantell, line 4: :+1: 
I should adopt this in NS as well, should clean up a few things

jsantell (approved):
LGTM!

PR #242, opened 2023-02-22
feat: Initial example of C integration

jsantell:
Need to figure out some text encoding stuff and clean ups, but this C snippet links and runs against libnoosphere

### DIFF
--- a/c/.gitignore
+++ b/c/.gitignore
@@ -1,0 +1,3 @@
+example/noosphere.h
+example/main.out
+example/main.o
--- a/c/README.md
+++ b/c/README.md
@@ -1,0 +1,16 @@
+# Noosphere C
+
+![License](https://img.shields.io/badge/license-MIT%2FApache--2.0-blue?label=License)
+[![Tests](https://img.shields.io/github/workflow/status/subconsciousnetwork/noosphere/Run%20test%20suite/main?label=Tests)](https://github.com/subconsciousnetwork/noosphere/actions/workflows/run_test_suite.yaml?query=branch%3Amain)
+
+An example/test case of binding to the noosphere static lib in C.
+
+## Running
+
+The included Makefile builds the `noosphere.h` header, and a debug build
+of `libnoosphere.a` if needed and links to a simple C test case. The test should build and run with either `gcc` or `clang`.
+
+```
+cd $NOOSPHERE_REPO/c/example
+make clean && make run
+```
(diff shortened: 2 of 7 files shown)


jsantell:
Now working. Some areas of improvements could be tying it into the larger build system:
* maybe lives in the noosphere crate?
* `make` is naive here as it's not rebuilding rust if it changes.
* Leak detection

Initial reactions to working with the raw FFI:
* Mostly copied the Swift tests, s/nil/NULL, and added semi-colons and it was smooth sailing other than remembering C trivia
* The generic wrapper types (`slice_ref_uint8_t`, `slice_boxed_char_ptr_t`) tripped me up, learning how safer_ffi maps Rust to C, initially criticizing the non-pointer args, but going back on that now. A small win here could be possibly renaming the generated types: `slice_boxed_char_ptr_t` -> `ns_string_array_t`, `slice_boxed_uint8_t` -> `ns_bytes` -- matching their corresponding `free` functions. But this challenge extends to all Rust-owned primitive types. Need to think more about this, wonder how other libraries handle this.
* The similarly named `slice_boxed_uint8_t` and `slice_ref_uint8_t` caused some confusion as well, with a slight difference in pointer types (ref being the `const` version) -- the "ref" type is for (from Rust's perspective) unowned data, used in args, and the "boxed" type is for return values. To simplify these intermediate types, maybe we can just use boxed for return values, and use two args (array, length) where needed (only `ns_sphere_fs_write` uses the ref type currently).
* Idiomatically, would love `NsSphereFs_t` -> `ns_sphere_fs_t`! Need to investigate feasibility of doing this in safer_ffi (meta issue: https://github.com/getditto/safer_ffi/issues/93)
* I had to link libmath (`-lm` clang/gcc arg) to get Noosphere linked, but that seems to be the only dependency

cdata (changes requested):
I love that we can add C to our retinue of supported languages. Some general feedback:

Can we put the example in a sub-folder? e.g., `$ROOT/c/example/main.c`

And, can we add a README.md at `$ROOT/c/README.md`? It doesn't have to be very long, but otherwise model it roughly after the other READMEs in our language folders.

Super cool :D

cdata:
Oh, also: please file an issue to get a test workflow in place for CI. It would be nice to be made aware of conspicuous support regressions.

jsantell:
* Added `HeaderTransformer` to transform generated header to C idiomatic types: `NsSphereFs_t` to `ns_sphere_fs_t`. Makes it possible to programmatically tag classes with correct Doxygen tags.
* Added explicit type casts between `uint8_t*` and `char*` to satiate clang (and actually try out both clang and gcc)
* Moved files around to make `c/` a root language directory like others
* Filed https://github.com/subconsciousnetwork/noosphere/issues/245

PR #249, opened 2023-02-28
fix: Limit delegated UCAN's lifetime to authorization token's lifetime where appropriate

### LINKED ISSUE #248: Intermittent: ns_test::it_is_thread_safe
The multithreaded NS test [has intermittently failed](https://github.com/subconsciousnetwork/noosphere/actions/runs/4285899776/jobs/7464766957)

```
thread 'it_is_thread_safe' panicked at 'called `Result::unwrap()` on an `Err` value: validation error', rust/noosphere-ns/tests/ns_test.rs:284:41
```

jsantell:
In some scenarios, both authorization and delegated tokens set their expiration time to be (NOW + 10000 years). If the delegated and auth tokens aren't created within the same timestamp-second, then the delegated token's lifetime exceeds its attenuation. This occurs intermittently (Fixes #248).

### DIFF
--- a/rust/noosphere-cli/src/native/commands/auth.rs
+++ b/rust/noosphere-cli/src/native/commands/auth.rs
@@ -5,7 +5,7 @@ use cid::Cid;
 use noosphere_core::{
     authority::{SphereAction, SphereReference},
     data::{CidKey, DelegationIpld, RevocationIpld},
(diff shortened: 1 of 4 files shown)


cdata (approved):
Nice catch!

Issue #243, opened 2023-02-24
C FFI for accessing petname address book


cdata:
Currently we are implementing generalized support for petname synchronization between clients and gateways (#214 and #215). Once that is in place, we should implement some C FFI surface area for accessing the address book and also publishing to the name system.

Issue #12, opened 2022-08-02
Pet name system


cdata:
Our plan at this time is to implement a generalized pet name system that enables users to write human-readable names instead of DIDs when addressing other spheres.

This is the tracking issue that covers progress towards our pet name system.

## Design notes

The proposed design for this system expects the following:

 - Every Sphere has an associated ed25519 key pair
 - Every Sphere revision can be referred to by a CID
 - A field in the Sphere maps human-readable names to DIDs (probably via a HAMT)
 - A DHT from which the CID of the latest revision of a Sphere can be resolved if the Sphere's public key is known

Spheres already have a key pair generated for them when they are first created. Spheres are already immutable data structures and can be looked up by CID.

The notion of a Sphere is sufficiently developed to encode a name-to-DID mapping, but the field needs to be added and incorporated with our versioning scheme. And, we have yet to implement a DHT.

## Errata

In our prototyping phase, we implemented a global, recursive name system using a CosmWasm smart contract. This was an interesting design, but ultimately we decided that a block chain is ill-suited to tracking mutable pointers. A single change could cost multiple pennies (denominated in USD), and recursive name resolution compounds the cost at each "hop" (where a hop is another contract invocation in this case). If every revision to a Sphere - where each individual revision could amount to a very small change - prompted a change to the pointer in the block chain, the cost to publish adds up quickly.

gordonbrander:
I think this may be done?

PR #253, opened 2023-03-07
Petname resolution and synchronization in spheres and gateways

### LINKED ISSUE #214: Store resolved names in a gateway's sphere
We have a viable stand-alone Noosphere nameserver (`orb-ns`) as of #207 . Now that we can resolve names in a more-or-less decoupled fashion, we should enhance the gateway to resolve names in a worker thread and then store them in the address book of its own sphere.

### LINKED ISSUE #215: Replicate resolved names from gateway sphere to user sphere
As of #214 , we will have implemented name resolution on the gateway side of the client<->gateway dyad. As a follow-on step, we should replicate newly resolved records for names to the client's address book. This should entail checking newly added records as the gateway's sphere is replicated to the client and propagating newly added or updated namerecords as they occur (one revision at a time).

### LINKED ISSUE #246: `additional_headers` in `SphereFs` should replace redundant headers by default
Currently, passing `additional_headers` to `SphereFs::write` or `SphereFs::save` will simply append headers to the existing list of headers in a memo. But, when we create that memo, it is typically "branched" from an existing memo. Branching entails cloning the memo verbatim, and retaining all headers except those that pertain to signing the content.

This means that if one wished to set a different value for a given custom header on save, the end result would be a pile-up of garbage headers from previous versions.

We should change this behavior so that custom headers are overwritten by default. And, maybe down the road we can consider allowing redundant headers from history to pile up.

cdata:
This change pertains to #12 and started as an effort to close the loop on end-to-end name publishing and resolution. Unfortunately, one thing lead to another and the change came to embody a significant refactor of how we read from and write to spheres in general.

## Publishing and resolving names

As of this change, we have implemented everything needed for a user to publish a version of their sphere to the Noosphere Name System, and for another user in the distance to resolve that published version by petname. We've also stubbed out the corresponding C FFI in preparation for #243.

Petname entries are added by a user to their own sphere. When the user sync's with their gateway, the gateway adopts the user's petname changes and kicks off a background job to resolve all the updated names from the Noosphere Name System. As those names are resolved, the gateway incorporates the resolved values into its address book. The next time a user sync's with the gateway after the resolutions are complete, the resolved values propagate back to their local device, and the sync strategy incorporates the new resolutions into the user's own address book.

Publishing happens on every push of the user's history to the sphere. It is technically optional to publish; it requires a specially crafted UCAN invocation, which is marked as optional in the body of the API request. But, we currently do it on every push anyway. When the user pushes, the gateway forwards the updated record to the Noosphere Name System in a background thread.

## The `SphereContext` refactor

The rationale for the refactor is that previously we had only one domain where we were performing a lot of user-driven reads and writes: `SphereFs`. We had factored things such that `SphereFs` did internal accounting of pending mutation state to a sphere. This immediately became a problem when we moved to add reading and writing APIs for a different domain (petnames), because `SphereFs` only dealt with the content part of a sphere.

The new factoring discards `SphereFs` and replaces it with a series of traits that ship the same API, but that are implemented directly for containers such as `Arc<SphereContext<_, _>>`. So, where previously we used one construct as a handle to a sphere - `SphereContext` - and a different one as a read/write API - `SphereFs` - we now do everything with `SphereContext`. New traits have been introduced to deal with petnames in a similar style to the traits that deal with content. And, all pending mutations are now "centrally" tracked by the `SphereContext` itself, so there is no risk of forking history by writing to two different parts of the sphere at the same time.

Fixes #246 
Fixes #214 
Fixes #215


rust/noosphere/src/ffi/petname.rs
```diff
@@ -0,0 +1,47 @@
...
+    petname: char_p::Ref<'_>,
+    error_out: Option<Out<'_, repr_c::Box<NsError>>>,
+) -> Option<char_p::Box> {
+    todo!();
+}
+
+#[ffi_export]
+/// Assign a DID to a petname. This will overwrite a petname entry if one already exists
+/// with the given name (and reset the resolved CID, if any).
+pub fn ns_sphere_petname_set(
```
jsantell, line 25: The tuple-nature of the address book (`(id, petname) => cid`) doesn't lead to obvious FFI patterns. We should document outcome if petname/did are NULL (implementation of course TBD). I don't think I'd expect the petname to be the "key" here, but changing DID or petname are both valid use cases. Wonder what it'd look like if e.g.

```c
links = ns_sphere_links_get();
for (uint8_t i = 0; i < links.len; i++) {
  ns_link* link = links.array[i];
  printf("name: %s", ns_link_get_name(link));
  printf("did: %s", ns_link_get_did(link));
  ns_link_set_name(link, "alice");
}
```

Though may cause similar synchronization issues as separation of Fs from Context.

rust/noosphere/src/ffi/petname.rs
```diff
@@ -0,0 +1,47 @@
...
+use safer_ffi::prelude::*;
+
+use crate::ffi::NsError;
+
+use super::NsNoosphereContext;
+
+#[ffi_export]
+/// Get the DID that is assigned to the provided petname; note that the DID is
+/// the ID of the sphere, but in order to read the sphere you must resolve the
+/// DID to a CID, which tells you the version of the sphere to read.
```
jsantell, line 12: A bit challenging to parse, separating the whats from the types; suggestion: "Get the identity as a DID that is assigned to the provided petname. To get the address associated with the petname, use `ns_sphere_petname_resolve()`."

The words "assigned to" and "associated with" suggests maybe we need a consistent name for this `(did, name)=>cid` object ("link"?) rather than "petname" being the core Thing here.

rust/noosphere/src/ffi/petname.rs
```diff
@@ -0,0 +1,47 @@
+#![allow(unused_variables)]
```
jsantell, line 1: Caution, bikeshedding here

rust/noosphere-api/Cargo.toml
```diff
@@ -20,6 +20,7 @@ readme = "README.md"
 
 [dependencies]
 anyhow = "^1"
+thiserror = "1.0.38"
```
jsantell, line 23: Should we use this going forward on decorating our own Errors?

jsantell (commented):
Looking great! some early comments so far

rust/noosphere-gateway/src/nns.rs
```diff
@@ -0,0 +1,319 @@
...
+    tokio::pin!(stream);
+
+    while let Some((name, address)) = stream.try_next().await? {
+        let last_known_record = address.last_known_record.clone();
+
+        let next_record =
+            match resolve_record(client.clone(), name.clone(), address.identity.clone()).await? {
+                Some(token) => {
+                    // TODO: Verify that the new value is the most recent value
+                    // TODO: Verify the proof chain of the new value
```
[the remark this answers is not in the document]
cdata: It is redundant but necessary. You may get a record from the DHT, but you may also get it from a peer, or a QR code, or....

rust/noosphere-api/Cargo.toml
cdata: Yah, we have used it in two places already

rust/noosphere-gateway/src/nns.rs
```diff
@@ -0,0 +1,319 @@
...
+                    Some(token)
+                }
+                None => {
+                    // TODO: Expire recorded value if we don't get an updated
+                    // record after some designated TTL
```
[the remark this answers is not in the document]
cdata: Yes!

rust/noosphere/src/ffi/petname.rs
cdata: You have the DID, yes. But then you need to resolve the DID to a CID. From there, the CID can be split into three parts: XID, YID and ZID. And with any combination of YID+ZID+DID or XID+YID+CID, you can derive the BID....

:D

rust/noosphere-core/src/view/sphere.rs
```diff
@@ -547,31 +549,70 @@ impl<S: BlockStore> Sphere<S> {
         ))
     }
 
-    /// Consume the [Sphere] and get a [Stream] that yields the [ChangelogIpld]
-    /// for content slugs at each version of the sphere.
-    pub fn into_link_changelog_stream(
+    /// NOTE: Steps backwards through history, excludes since
```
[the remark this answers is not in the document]
cdata: Yah, my bad. The note was for myself :sweat_smile:

rust/noosphere/src/ffi/petname.rs
cdata: Note that under the hood, names (like content) are contained by a HAMT. This data structure is convenient (and necessary) because it allows us operate on the data with map semantics, while also allowing us to avoid loading in the entire data structure just to read one key.

The implication here is that loading in the entire set of names could be quite expensive as a general purpose operation, especially if we are working on one key at a time. It's expensive in terms of memory allocation and it's expensive in terms of storage access because the key space of the "map" is effectively sharded across many blocks.

That having been said, even in cases where synchronization _is not_ an issue, we would still want to mediate access to the list of petnames via FFI. And, on the Rust side we implement iteration over all petnames/slugs/etc. as `Stream`s - which are effectively async generators - so that we can side-step the pathological case of loading the entire key space into memory at once.

So, I guess the takeaway here is that we need a FFI to access the full space of petnames.

rust/noosphere/src/ffi/petname.rs
cdata: I have a separate branch of work going for filling out the C FFI. I'll try to address your feedback in that change.

rust/noosphere-core/src/view/sphere.rs
```diff
@@ -547,31 +549,70 @@ impl<S: BlockStore> Sphere<S> {
...
+                let changelog = names.try_load_changelog().await?;
+
+                yield (cid, changelog);
+            }
+        }
+    }
+
+    /// Consume the [Sphere] and get a [Stream] that yields the [ChangelogIpld]
+    /// for content slugs at each version of the sphere.
+    pub fn into_link_changelog_stream(
```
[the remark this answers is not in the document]
cdata: We might have to manifest this as a macro helper. Changes here refer to the change in a `VersionedMap` are typed `Changelog<K, V>`, so the method invocation needs to imply key and value types in addition to the field.

rust/noosphere/src/ffi/petname.rs
jsantell: Well explained, thanks!

rust/noosphere-sphere/src/sync/gateway.rs
```diff
@@ -151,73 +202,100 @@ where
...
+
+    async fn adopt_names(
+        &self,
+        context: &mut H,
+        updated_names: BTreeMap<String, AddressIpld>,
+    ) -> Result<Option<Cid>> {
+        if updated_names.len() == 0 {
+            return Ok(None);
+        }
+        println!(
```
jsantell, line 259: No `println` in `noosphere` imo, this should be rendered in `orb` if needed.

There's plenty of existing println here though, let's kick off our rules to writing to stdout in a separate issue

rust/noosphere-sphere/src/sync/gateway.rs
```diff
@@ -151,73 +202,100 @@ where
...
+            AddressIpld {
+                identity,
+                last_known_record,
+            },
+        ) in updated_names.into_iter()
+        {
+            if let Some(jwt) = last_known_record {
+                context.adopt_petname(&name, &identity, &jwt).await?;
+            } else {
+                warn!("UH OH");
```
jsantell, line 275: :fearful:

rust/noosphere-sphere/src/sync/gateway.rs
```diff
@@ -231,22 +309,51 @@ where
...
+            .witnessed_by(&authorization)
+            .claiming_capability(&Capability {
+                with: With::Resource {
+                    kind: Resource::Scoped(SphereReference {
+                        did: local_sphere_identity.to_string(),
+                    }),
+                },
+                can: SphereAction::Publish,
+            })
+            .with_lifetime(120)
```
jsantell, line 336: Wouldn't this record expire in 2 minutes? We should have some constant, default lifetime duration somewhere

rust/noosphere-sphere/src/helpers.rs
```diff
@@ -0,0 +1,55 @@
+// #![cfg(any(test, doc, feature = "helpers"))]
```
jsantell, line 1: nit: remove

rust/noosphere-sphere/src/helpers.rs
```diff
@@ -0,0 +1,55 @@
...
+/// Access levels available when simulating a [SphereContext]
+pub enum SimulationAccess {
+    Readonly,
+    ReadWrite,
+}
+
+/// Create a temporary, non-persisted [SphereContext] that tracks usage
+/// internally. This is intended for use in docs and tests, and should otherwise
+/// be ignored. When creating the simulated [SphereContext], you can pass a
+/// [SimulationAccess] to control the kind of access the emphemeral credentials
```
jsantell, line 26: nit: "ephemeral"

rust/noosphere-ns/src/helpers.rs
```diff
@@ -0,0 +1,86 @@
...
+    for _ in 0..ns_count {
+        let ns_data =
+            generate_name_system(&mut store, &bootstrap_addresses, validator.clone()).await?;
+        name_systems.push(ns_data);
+    }
+
+    Ok((bootstrap_node, store, name_systems))
+}
+
+pub fn generate_default_listening_address() -> Multiaddr {
```
jsantell, line 84: This logic was originally more complex, and has since simplified due to "0" ports and other changes, and could now simply be a static value.

I think that this function should *not* be publically exposed due to not wanting to prioritize 127.0.0.1 vs 0.0.0.0. In lieu of that, I think this function should remain in `dht_test.rs`, or simply inlined where needed.

rust/noosphere-ns/src/lib.rs
```diff
@@ -14,13 +14,15 @@ mod records;
 pub mod utils;
 mod validator;
 
+pub mod helpers;
```
jsantell, line 17: There's `mod utils` which serves a similar purpose, though `helpers` specifically is for tests -- maybe can consolidate these into a single module, or have helper module (or functions) gated on `cfg(test)`. I don't think `NsData` should be exposed by default.

rust/noosphere-ns/src/dht/processor.rs
```diff
@@ -68,12 +68,14 @@ where
         keypair: &libp2p::identity::Keypair,
         peer_id: PeerId,
         validator: Option<V>,
-        config: DHTConfig,
-        processor: DHTMessageProcessor,
-    ) -> Result<tokio::task::JoinHandle<Result<(), DHTError>>, DHTError> {
+        config: DhtConfig,
+        processor: DhtMessageProcessor,
+    ) -> Result<tokio::task::JoinHandle<Result<(), DhtError>>, DhtError> {
+        // debug!("SPAWNING WITH VALIDATOR: {:?}", validator);
```
jsantell, line 74: nit: comment

rust/noosphere-ns/src/name_system.rs
```diff
@@ -39,24 +39,24 @@ lazy_static! {
...
 }
 
 impl NameSystem {
-    pub fn new<S: Storage + 'static, K: DHTKeyMaterial>(
+    pub fn new<K: DhtKeyMaterial, V: RecordValidator + 'static>(
         key_material: &K,
-        store: SphereDb<S>,
-        dht_config: DHTConfig,
+        dht_config: DhtConfig,
+        validator: Option<V>,
```
jsantell, line 56: I don't like a configurable/optional top-level validator (wanting a network of nodes to have relatively similar/consistent validators), but I don't have any other ideas in order to support this (and more pushing and pulling between static DhtConfig values and living instances. I'm glad you've spent some time in this module and the `Dht*` counterparts for future discussions on some of the cleanup challenges ahead!

rust/noosphere-ns/src/records.rs
```diff
@@ -12,7 +12,7 @@ use std::{convert::TryFrom, str, str::FromStr};
 use ucan::{builder::UcanBuilder, crypto::KeyMaterial};
 use ucan::{chain::ProofChain, crypto::did::DidParser, Ucan};
 
-/// An [NSRecord] is the internal representation of a mapping from a
+/// An [NsRecord] is the internal representation of a mapping from a
```
jsantell, line 15: I'm not ready to talk about how `NsRecord`/`NSRecord` is the only struct prefixed by `NS`  :upside_down_face:

rust/noosphere-ns/src/helpers.rs
```diff
@@ -0,0 +1,86 @@
...
+    let delegate_capability = generate_capability(&sphere_id);
+    let delegation = UcanBuilder::default()
+        .issued_by(&sphere_key)
+        .for_audience(&owner_id)
+        .with_lifetime(SPHERE_LIFETIME)
+        .claiming_capability(&delegate_capability)
+        .build()?
+        .sign()
+        .await?;
+    let _ = store.write_token(&delegation.encode()?).await?;
```
jsantell, line 41: Kind of a bummer that `store` is being passed around these utils only for this side-stepping-Ipfs hack here, now that `store` is internal to the `validator`

rust/noosphere-ns/src/validator.rs
```diff
@@ -22,17 +22,31 @@ where
     }
 }
 
+impl<S> Clone for Validator<S>
```
jsantell, line 25: Should this also implement Display like the AllowAllValidator?

jsantell (approved):
Handful of comments with a closer eye on the NS changes, and some small things like comments. Excellent, gigantic work!!

rust/noosphere-ns/src/validator.rs
cdata: Ah, `Display` is vestigial, I can remove that implementation

rust/noosphere-ns/src/records.rs
cdata: :sweat_smile:

rust/noosphere-cli/src/native/commands/sphere.rs
```diff
@@ -109,6 +107,7 @@ Type or paste the code here and press enter:"#,
         .authorized_by(Some(&Authorization::Cid(cid)))
         .build()
         .await?;
+    debug!("B");
```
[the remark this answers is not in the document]
cdata: Explanation: b

rust/noosphere-sphere/src/sync/gateway.rs
cdata: SG, although worth noting that at this time we have no expiration of name records (at least, not one that we enforce).

Already filed https://github.com/subconsciousnetwork/noosphere/issues/259 to discuss

rust/noosphere-ns/src/lib.rs
cdata: I tried gating this on `cfg(test)` and then on a feature flag, but there is no satisfying way that I could find where this stuff is _only_ exposed to tests in both integration tests and other crates. Open to suggestions.

rust/noosphere-ns/src/name_system.rs
cdata: Pasting my related comment from Discord:

> I'm sympathetic to your position on configurable Validator. Ultimately, I don't think orb-ns will offer a configuration here. Nor would orb. And, it may prove to be an anti-pattern altogether.

And filed https://github.com/subconsciousnetwork/noosphere/issues/268

cdata:
Alright folks, we're gonna land this one.

PR #254, opened 2023-03-08
Reconfigure rust module dependencies

jsantell:
* Update graphviz rendering based off of #253 and these changes
* Move `IpfsStorage` trait to `noosphere-ipfs` from `noosphere-storage`. Storage depending on `ipfs` (which in turn, in tests uses core/storage) lead to cyclical dependencies and is a helpful code smell considering `storage` is a load-bearing module.


cdata (approved):
Awesome, thanks for untangling this!

PR #270, opened 2023-03-10
fix: FFI header generation

jsantell:
Use an ordered BTreeMap to replace class token keys so that class names that are subsets of other class names are replaced appropriately.


cdata (approved):
Nice fix! :beers:

PR #271, opened 2023-03-13
feat!: Implement C FFI for petname management

### LINKED ISSUE #243: C FFI for accessing petname address book

cdata:
This is a follow-up to #253 that implements that C FFI surface for petname support. As part of a general transition away from the deprecated `SphereFs` and towards `SphereContext` + traits, this also tweaks the signatures of FFI methods in a (hopefully) non-breaking way to indicate that `SphereFs` is no longer a thing.

Note that this change does not include validation of petname records. A change is already in-progress to implement validation in the gateway.

Fixes #243

### DIFF
--- a/rust/noosphere-sphere/src/walker.rs
+++ b/rust/noosphere-sphere/src/walker.rs
@@ -1,6 +1,6 @@
 use anyhow::Result;
 use cid::Cid;
-use noosphere_core::data::MapOperation;
+use noosphere_core::data::{AddressIpld, MapOperation};
 use std::{collections::BTreeSet, marker::PhantomData};
 
 use async_stream::try_stream;
@@ -9,15 +9,17 @@ use tokio::io::AsyncRead;
 use tokio_stream::{Stream, StreamExt};
 use ucan::crypto::KeyMaterial;
 
-use crate::{internal::SphereContextInternal, HasSphereContext};
+use crate::{
+    content::{SphereContentRead, SphereFile},
+    internal::SphereContextInternal,
+    HasSphereContext, SpherePetnameRead,
+};
 
-use super::{SphereContentRead, SphereFile};
-
-/// A [SphereContentWalker] makes it possible to convert anything that
-/// implements [HasSphereContext] into an async [Stream] over sphere content,
-/// allowing incremental iteration over both the breadth of content at any
-/// version, or the depth of changes over a range of history.
-pub struct SphereContentWalker<H, K, S>
+/// A [SphereWalker] makes it possible to convert anything that implements
+/// [HasSphereContext] into an async [Stream] over sphere content, allowing
+/// incremental iteration over both the breadth of content at any version, or
+/// the depth of changes over a range of history.
+pub struct SphereWalker<H, K, S>
 where
     H: HasSphereContext<K, S>,
     K: KeyMaterial + Clone + 'static,
@@ -28,14 +30,14 @@ where
     storage: PhantomData<S>,
 }
 
-impl<H, K, S> From<H> for SphereContentWalker<H, K, S>
+impl<H, K, S> From<H> for SphereWalker<H, K, S>
 where
     H: HasSphereContext<K, S>,
     K: KeyMaterial + Clone + 'static,
     S: Storage + 'static,
 {
     fn from(has_sphere_context: H) -> Self {
-        SphereContentWalker {
+        SphereWalker {
             has_sphere_context,
             key: Default::default(),
             storage: Default::default(),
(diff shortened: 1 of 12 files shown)


rust/noosphere/src/ffi/petname.rs
```diff
@@ -1,47 +1,173 @@
...
+        })
+    })
 }
 
 #[ffi_export]
-/// Assign a DID to a petname. This will overwrite a petname entry if one already exists
-/// with the given name (and reset the resolved CID, if any).
+/// Assign a petname to a sphere identity (a DID). This will overwrite the
+/// petname with the new sphere identity if it has been set previously. When a
+/// new address is assigned to a petname, the petname will be reset to an
```
jsantell, line 41: nit: s/address/identity to be consistent with first sentence

rust/noosphere/src/ffi/petname.rs
```diff
@@ -1,47 +1,173 @@
...
+/// unresolved state. Make sure to invoke sync after setting new petnames, so
+/// that new sphere identities can be resolved by the gateway (if one is
+/// configured).
```
jsantell, line 44: nit: In `ns_sphere_petname_get` and `ns_sphere_petname_set` (line 3), a sphere identity/DID is *assigned to* a petname. In `ns_sphere_petname_set` line 1, a petname is *assigned to* an identity/DID. While I do think of the structure as assigning a petname to identity, the inverse relationship is true and more common in comments (e.g. that the petname, not DID, is the key)

rust/noosphere/src/ffi/petname.rs
```diff
@@ -1,47 +1,173 @@
...
+    })
 }
 
 #[ffi_export]
-/// Assign a DID to a petname. This will overwrite a petname entry if one already exists
-/// with the given name (and reset the resolved CID, if any).
+/// Assign a petname to a sphere identity (a DID). This will overwrite the
+/// petname with the new sphere identity if it has been set previously. When a
+/// new address is assigned to a petname, the petname will be reset to an
+/// unresolved state. Make sure to invoke sync after setting new petnames, so
```
jsantell, line 42: What happens if an "unresolved state" is propagated in a sphere? This reads as "undefined behaviour"

rust/noosphere/src/ffi/petname.rs
```diff
@@ -1,47 +1,173 @@
...
+        })
+    });
 }
 
 #[ffi_export]
-/// Resolve a configured petname to a sphere version (a CID), via the DID that
-/// has been assigned to it. The returned value is a UTF-8, base64-encoded CIDv1
-/// string. If no DID has been assigned to the given petname, no value will be
-/// resolved.
+/// Resolve a configured petname, using the sphere idenitty - a DID - to
```
jsantell, line 65: nit: s/idenitty/identity

rust/noosphere/src/ffi/petname.rs
```diff
@@ -1,47 +1,173 @@
...
+/// determine a link - a CID - that is associated with it. The returned link is
+/// a UTF-8, base64-encoded CIDv1 string that may be used to resolve data from
+/// the IPFS content space. Note that this call will produce an error if no
+/// address has been assigned to the given petname.
```
jsantell, line 69: Unclear what scenario produces an error, as "address" is unclear -- "address" could be "identity" (referring to a bound DID) or "link" (referring to a NS-resolved CID), and IIUC, two different scenarios could result in this. Worth renaming "address" to either "identity" or "link" (IMO, I think most of the comments are pretty consistent with it), and including all scenarios that produce error (which may be if either "link" or "identity" are absent).

rust/noosphere/src/ffi/petname.rs
```diff
@@ -1,47 +1,173 @@
...
 pub fn ns_sphere_petname_resolve(
-    noosphere: &NsNoosphereContext,
-    sphere_identity: char_p::Ref<'_>,
+    noosphere: &NsNoosphere,
+    sphere: &NsSphere,
```
jsantell, line 72: These consistent arg changes throughout are great :+1:

rust/noosphere/src/ffi/petname.rs
```diff
@@ -1,47 +1,173 @@
...
+        None => Vec::new(),
+    }
+    .into_boxed_slice()
+    .into()
+}
+
+#[ffi_export]
+/// Get an array of all of the petnames that changed in a given sphere since a
+/// given revision of that sphere (excluding the given revision). The revision
+/// should be provided as a UTF-8 base64-encoded CIDv1 string. If no revision is
```
jsantell, line 125: It's great, verbose, and necessary, to document the string types used on this interface, e.g. "UTF-8 base64-encoded CIDv1 string", wonder if we can consolidate that somehow in documentation without some extra opaque type in consuming platforms.

rust/noosphere/src/ffi/petname.rs
```diff
@@ -1,47 +1,173 @@
...
+/// provided, the entire history will be considered (back to and including the
+/// first revision).
+///
+/// Note that a petname change may mean the petname was added, updated or
+/// removed. Also note that multiple changes to the same petname will be reduced
+/// to a single entry in the array that is returned.
+///
+/// A petname will also be considered changed if it goes from an unresolved
+/// state to a resolved state.
```
jsantell, line 134: Ah I see, "resolved" and "unresolved" are explicit states of a petname

rust/noosphere/src/ffi/petname.rs
```diff
@@ -1,47 +1,173 @@
...
+#[ffi_export]
+/// Get an array of all of the petnames that changed in a given sphere since a
+/// given revision of that sphere (excluding the given revision). The revision
+/// should be provided as a UTF-8 base64-encoded CIDv1 string. If no revision is
+/// provided, the entire history will be considered (back to and including the
+/// first revision).
+///
+/// Note that a petname change may mean the petname was added, updated or
+/// removed. Also note that multiple changes to the same petname will be reduced
+/// to a single entry in the array that is returned.
```
jsantell, line 131: Good clarification here

swift/Tests/SwiftNoosphereTests/NoosphereTests.swift
```diff
@@ -407,4 +407,189 @@ final class NoosphereTests: XCTestCase {
...
+
+        let sphere_receipt = ns_sphere_create(noosphere, "bob", nil)
+
+        let sphere_identity_ptr = ns_sphere_receipt_identity(sphere_receipt, nil)
+        let sphere = ns_sphere_fs_open(noosphere, sphere_identity_ptr, nil)
+        
+        ns_string_free(sphere_identity_ptr)
+        ns_sphere_receipt_free(sphere_receipt)
+        
+        ns_sphere_petname_set(noosphere, sphere, "alice", "did:key:alice", nil)
```
jsantell, line 424: Future Thoughts: Wonder if we can bake the noosphere context into the sphere to simplify API in the future by removing the `noosphere` arg when a `sphere` has been provided (and "opened" via a context):

```c
ns_sphere_t sphere = ns_sphere_fs_open(noosphere, "did:key:bob",  NULL);
ns_sphere_petname_set(sphere, "alice", "did:key:alice", NULL);
```

swift/Tests/SwiftNoosphereTests/NoosphereTests.swift
```diff
@@ -407,4 +407,189 @@ final class NoosphereTests: XCTestCase {
...
+        
+        assert(change_count == expected_changes.count)
+        
+        var pointer = changes.ptr!
+        
+        for i in 0..<change_count {
+            let petname = String.init(cString: pointer.pointee!)
+            
+            print("Changed petname:", petname)
+            assert(expected_changes[i] == petname)
```
jsantell, line 584: If the FFI only exposes *which* petnames have been changed, how does the use case here work in e.g. Subconscious?

* Query noosphere for changed petnames
* Compare changed petnames to what's stored in e.g. sqlite, inferring the operation (add, remove, change) by comparing `ns_sphere_petname_get` or resolve from the stored value?

jsantell (approved):
LGTM! The petname structure has been a challenging design and great to see this entering FFI-land. Lots of design questions in the future w/r/t petnames, so comments here are mostly pedantic bikeshedding over terms that we can save for another day

rust/noosphere/src/ffi/petname.rs
cdata: A name / DID pair resolve to a link, just like in DNS a domain name resolves to the value side of a DNS record. If it is resolved, you have a link. If it is unresolved, you don't have a link.

rust/noosphere/src/ffi/petname.rs
cdata: An address can never be link, this is just a typo where the word address should be identity.

swift/Tests/SwiftNoosphereTests/NoosphereTests.swift
cdata: It's possible, file an issue?

swift/Tests/SwiftNoosphereTests/NoosphereTests.swift
cdata: Yeah, it's a bit unclear in the "remove" case, I guess. "remove" and "changed but now unresolved" are ambiguous. I'll see if I can address that somehow.

rust/noosphere/src/ffi/petname.rs
cdata: I'm happy to standardize on whatever order makes the most sense.

bfollington: My intuition diverges here depending on whether you're thinking from the UI layer or the protocol layer. 

Within Noosphere it seems correct to say that you _assign_ a petname as a label for a DID, where-as in Subconscious it feels natural to say you create an address book entry by entering a petname and _assigning_ a DID to it (though we can flip that around through UI design if we want to).

Given that this is a Noosphere-level abstraction I'd lean towards "a petname is assigned to a DID", personally.

swift/Tests/SwiftNoosphereTests/NoosphereTests.swift
```diff
@@ -407,4 +407,189 @@ final class NoosphereTests: XCTestCase {
...
+        let identity = String.init(cString: identity_ptr!)
+        
+        assert(identity == "did:key:alice")
+        
+        ns_string_free(identity_ptr)
+        ns_sphere_fs_free(sphere)
+        ns_free(noosphere)
+    }
+
+    func testGetAllPetnamesInASphere() throws {
```
bfollington, line 437: This test (and the next one) are _extremely_ helpful for understanding the consumption side of the story 🙏

rust/noosphere-sphere/src/walker.rs
```diff
@@ -28,31 +30,153 @@ where
...
+
+    /// Get a [BTreeSet] whose members are all the petnames that have addresses
+    /// as of this version of the sphere. Note that the full space of names may
+    /// be very large; for a more space-efficient approach, use
+    /// [SphereWalker::petname_stream] to incrementally access all petnames in
+    /// the sphere.
+    ///
+    /// This method is forgiving of missing or corrupted data, and will yield an
+    /// incomplete set of names in the case that some or all names are not able
+    /// to be accessed.
```
bfollington, line 109: Question: is there a specific cause of missing data / corruption in mind here that would result in failing to read a petname?

cdata: A very common cause may be: you haven't fully replicated the sphere you are reading from yet. Ideally, this is a protocol detail that should not leak to the caller, but in practice I'm certain we will hit this from time to time.

IPLD is nice because it enables us to incrementally load a sphere when we want to read a narrow slice of it. But, one drawback of this is: just because you have part of a sphere doesn't mean you have the whole sphere. If you load part of the address book (say, enough to read the key space, but not all the values), and then go offline, you may not be able to read parts of the address book that you are iterating over.

There is probably a UX to be explored where we indicate to the reader that the sphere is in an indeterminate state (due to only being partially available), to better communicate what is happening in cases like this.

jsantell (approved):

swift/Tests/SwiftNoosphereTests/NoosphereTests.swift
cdata: Added `ns_sphere_petname_is_set` to resolve the ambiguity

swift/Tests/SwiftNoosphereTests/NoosphereTests.swift
cdata: @jsantell filed #272

rust/noosphere/src/ffi/petname.rs
cdata: We may be able to hoist some of this to module-level with sufficient work on the header transformer?

jsantell: Aye.. I'll file something more specific about "class" documentation that this could cover

Issue #269, opened 2023-03-10
Enable traversal from one sphere to the next


cdata:
When we store sphere data for spheres that belong to other authors besides the configured author - where "author" is essentially a key holder - we store it in the same storage as the author's own sphere. So, for a given author accessing a third-party sphere to reliably access the blocks that have been previously fetched, we need a handle to their credentials as well as the backing storage layer.

Sphere "traversal" is the process by which the author of one sphere accesses another sphere via the context of their own sphere.

In order to implement this feature, we need to perform the following (roughly):

 1. Initialize the origin `SphereContext` for the sphere the author is starting from
 2. Begin traversal by specifying the DID of the destination sphere
 3. If a gateway is configured, create a pending request to the gateway to replicate the destination sphere locally
 4. If the destination sphere data is available locally, immediately resolve it; otherwise, block on the pending request to the gateway before resolving it
 5. Create a `SphereContext` for the destination sphere configured with the same author and storage backend as the origin `SphereContext`

Issue #267, opened 2023-03-09
Support `IpfsStorage` if an IPFS API provided in gateway


jsantell:

jsantell:
Updating this to focus on validation in the name system (mostly tests and infra), with #257 focusing on validating records in the gateway.

jsantell:
disabling of validation in #356 will need undone

cdata:
Fixed in #378

PR #278, opened 2023-03-16
feat: Revise links and gateway

cdata:
This change is a precursor to #269 . It represents two important revisions that clean things up to prepare for sphere traversal.

 1. The gateway is now generic to the storage type and sphere context provider it supports
 2. Sphere link values are now `MemoIpld` instead of `Cid`

(1) enables us to pass in an `IpfsStorage`-wrapped storage to the gateway without changing the underlying implementation to be aware of it.

(2) aligns links with the way we store other constructs (like petnames and authorizations), and enables us to clean up a gross corner of how we determine what must be bundled for a given sphere revision to be transferred over the network.

### DIFF
--- a/rust/noosphere-core/src/data/bundle.rs
+++ b/rust/noosphere-core/src/data/bundle.rs
@@ -6,20 +6,17 @@ use cid::Cid;
 
 use futures::{pin_mut, StreamExt};
 use libipld_cbor::DagCborCodec;
-use libipld_core::{raw::RawCodec, serde::to_ipld};
+use libipld_core::raw::RawCodec;
 use noosphere_storage::{block_deserialize, block_serialize, BlockStore};
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 
 use crate::{
     data::{
-        BodyChunkIpld, ChangelogIpld, ContentType, Header, LinksIpld, MapOperation, MemoIpld,
-        SphereIpld, VersionedMapIpld,
+        AddressIpld, AllowedIpld, AuthorityIpld, BodyChunkIpld, ChangelogIpld, ContentType,
+        DelegationIpld, Header, LinksIpld, MapOperation, MemoIpld, NamesIpld, RevocationIpld,
+        RevokedIpld, SphereIpld, VersionedMapIpld, VersionedMapKey, VersionedMapValue,
     },
     view::Timeslice,
-};
-
-use super::{
-    AllowedIpld, AuthorityIpld, NamesIpld, RevokedIpld, VersionedMapKey, VersionedMapValue,
 };
 
 // TODO: This should maybe only collect CIDs, and then streaming-serialize to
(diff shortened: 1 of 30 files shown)


rust/noosphere-core/src/data/bundle.rs
```diff
@@ -182,49 +179,26 @@ impl TryBundle for BodyChunkIpld {
...
+    V: VersionedMapValue + TryBundle,
 {
     async fn try_extend_bundle_with_cid<S: BlockStore>(
         cid: &Cid,
         bundle: &mut Bundle,
         store: &S,
     ) -> Result<()> {
         let bytes = store.require_block(cid).await?;
-        let mut cids = Vec::new();
+        // let mut cids = Vec::new();
```
jsantell, line 190: nit: comment

rust/noosphere-core/src/data/bundle.rs
```diff
@@ -182,49 +179,26 @@ impl TryBundle for BodyChunkIpld {
 impl<K, V> TryBundle for ChangelogIpld<MapOperation<K, V>>
 where
     K: VersionedMapKey,
-    V: VersionedMapValue,
+    V: VersionedMapValue + TryBundle,
 {
     async fn try_extend_bundle_with_cid<S: BlockStore>(
```
jsantell, line 184: consider: removal of `try_` prefix, like the other `noosphere` (IIRC?) changes

rust/noosphere-gateway/src/nns.rs
```diff
@@ -33,39 +32,36 @@ pub enum NameSystemConfiguration {
...
     },
     Publish {
-        context: Arc<Mutex<SphereContext<K, S>>>,
+        context: C,
         record: Jwt,
     },
 }
 
-pub fn start_name_system<K, S>(
+pub fn start_name_system<H, K, S>(
```
jsantell, line 59: nit consider throughout: consistently use the same generic (e.g. `C` or `Ctx` maybe) for bounded `H: HasMutableSphereContext` and unbounded (`enum NameSystemJob<C>`) forms, and consider adding bounds to unbounded (`OnDemandNameResolver<C>`) forms

jsantell (dismissed):
Very nice! Necessary for #267 as well I believe

rust/noosphere-gateway/src/nns.rs
cdata: I'll standardize the letter, but adding bounds to those other forms is very verbose (because the extra generics need to be expressed in terms of `PhantomData` and/or unusable enum variants). Also, it's unnecessary because the bounds are enforced implicitly (e.g., but the associated `Sender`).

jsantell: I see!

cdata: Every `H` is now a `C`.

jsantell (approved):

rust/noosphere-core/src/data/bundle.rs
cdata: `try_` removed

Issue #232, opened 2023-02-16
[orb]: Error: No value found for 'user_key_name'


jsantell:
Calling some commands `orb config get counterpart`, `orb serve`, etc. results in the error `Error: No value found for 'user_key_name'`

jsantell:
This occurs when firing sphere-specific commands outside of a sphere directory, e.g. `orb key list`

jsantell:
Fixed via #280

PR #280, opened 2023-03-17
feat: Improve "orb" error messaging when sphere directory is expected and does not exist

jsantell:
...and other orb papercuts.
* Prevents ambitious creation of an uninitialized ".sphere" dir that lacks a "user_key_name", fixing #232. (thus causing many phantom `.sphere` dirs)
* Updates and constrains "clap" to match "noosphere-ns".
* Adds a hidden alias of "ls" to "list" in "orb" e.g. "orb key ls".
* Hoists tracing utility to native::main


cdata (approved):
Nice!

PR #282, opened 2023-03-20
chore: upgrade safer_ffi to 0.1.0-rc1

jsantell:
* Current version of safer ffi (0.0.10) is 10 months out of date from the RC 0.1.0-rc1, bringing a handful of new features.
 * Initial Doxygen-style `@class` annotation output -- lacking actual docs content, but stubbed.
 * Upgraded safer_ffi has built-in class renaming; removing custom header transformation logic.
 * `safer_ffi::bool` no longer exists; C doesn't have booleans, `u8` seems appropriate for `ns_sphere_petname_is_set`.
 * Some new logic w/r/t to out variables in latest version, but doesn't seem like we need to change anything yet on our end.


cdata (approved):
Awesome, thanks for coordinating with the maintainer on this!

PR #285, opened 2023-03-23
chore: additional work on FFI documentation.

jsantell:
* Expand @\class and @\memberof Doxygen relationships amongst FFI functions.
* Refine some FFI docs for consistent language and adapted to Doxygen's "summary" and "detailed" views (via newlines).
* Fix up rustdoc lints.

### DIFF
--- a/rust/noosphere/src/ffi/headers.rs
+++ b/rust/noosphere/src/ffi/headers.rs
@@ -3,8 +3,9 @@ use safer_ffi::prelude::*;
 #[derive_ReprC(rename = "ns_headers")]
 #[repr(opaque)]
 /// @class ns_headers_t
+/// An opaque struct representing name/value headers.
 ///
-/// TBD
+/// Headers are used in ns_sphere_file_t to assign metadata, like content type.
 pub struct NsHeaders {
     inner: Vec<(String, String)>,
 }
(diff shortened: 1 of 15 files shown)


rust/noosphere/src/ffi/headers.rs
```diff
@@ -3,8 +3,18 @@ use safer_ffi::prelude::*;
 #[derive_ReprC(rename = "ns_headers")]
 #[repr(opaque)]
 /// @class ns_headers_t
+/// An opaque struct representing name/value headers.
 ///
-/// TBD
+/// Headers are used in ns_sphere_file_t to assign metadata, like content type.
```
jsantell, line 8: An example of "class" examples. As we only have one big example for now, unclear if having individual (untested??) examples is better. Can we run non-rust tests as a part of rustdocs? Hmm

rust/noosphere/src/ffi/noosphere.rs
```diff
@@ -98,26 +91,29 @@ pub fn ns_initialize(
 }
 
 #[ffi_export]
-/// De-allocate a [NoosphereContext]. Note that this will also drop every
-/// [SphereContext] that remains active within the [NoosphereContext].
+/// @memberof ns_noosphere_t
+/// Deallocate a ns_noosphere_t instance.
+///
+/// Note that this will also deallocate every ns_sphere_t that remains
```
jsantell, line 97: Notably, this extra new line here creates the fully detailed function on the docs page with extended information. We could annotate other data e.g. @\param, @\return for more clarity as well, which would also create the "full" function def upon generation.

jsantell (commented):
Some followups:
* Annotating safer_ffi's built-in wrapper types (`slice_boxed_char_ptr`). The generated rust comments from here are a little inscrutable.
* Refining some global functions into a phantom "class", as of now they're a bit hidden e.g. `ns_string_free()`, accessible only from the `noosphere.h` "file" "global functions".
* Investigate if we can test C functions in rustdocs, and other ways we can integrate examples in documentation
* Add links to landing page so that JS isn't required to navigate docs.

rust/noosphere/src/key/insecure.rs
```diff
@@ -56,7 +56,7 @@ impl InsecureKeyStorage {
     /// but for now we aren't certain if we can depend on the ability to
     /// enumerate keys in general WebAuthn cases.
     ///
-    /// See: https://www.w3.org/TR/webauthn-3/#client-side-discoverable-credential
+    /// See: <https://www.w3.org/TR/webauthn-3/#client-side-discoverable-credential>
```
cdata, line 59: How do the angle brackets help in this case?

rust/noosphere/src/ffi/headers.rs
cdata: I would actually prefer not to have code examples in docs for the very reason that we can't easily keep them up to date. Better to have actual tests!

jsantell: agreed, examples running against CI way preferable at this point, will remove this

jsantell:
@cdata updated & rebased

cdata (approved):
Awesome work!

Issue #268, opened 2023-03-10
Remove configurable validation from NNS


cdata:
As of #253 it is possible for us to inject validation logic from the entrypoint of the Noosphere Name System. However, it is quite possible that this is an anti-pattern (even though it makes building early tests very convenient). In the end, we want all implementations of the name system to rely on prescribed validation logic, so we probably don't want validation to be pluggable at all.

PR #290, opened 2023-03-24
feat: validate proofs over IPFS of NsRecords in the gateway

### LINKED ISSUE #257: Verify proof chain for name record adoption candidates
As of #253 , we will have end-to-end name publishing and resolving. However, only the name resolvers currently do any validation of name records. Name records must also be validated at the time that they are adopted by a sphere (this includes user spheres and Geist spheres), because although most names may be coming from the name system, many may come from other sources such as other users, digital or physical links (like QR codes) or even manual entry. So, proof chains must be verified at adoption time to ensure their correctness.

### LINKED ISSUE #267: Support `IpfsStorage` if an IPFS API provided in gateway

jsantell:
Fixes #267. Fixes #257.

* Gate `noosphere-ipfs` storage capabilities behind feature "storage".
* Extend `noosphere_ipfs:rIpfsStore` with UcanStore trait.
* Fix `noosphere_ipfs::KuboClient` to be codec-aware when fetching blocks.
* Simplify components in `noosphere_ns`:
  * Use UcanJwtStore trait as input for NameSystem rather than Validator. Fixes https://github.com/subconsciousnetwork/noosphere/issues/268.
  * Reduce external usage on DidParser in NsRecord validation.
  * Decouple NameSystem peering helpers from authorization/sphere testing helpers.

--

Currently:
* Initial extending of NS storage with IpfsStorage for validation worked so well, it breaks tests! remaining failures are in peer_to_peer.rs
* big cleanups in NS as a prereq, leveraging new experience and knowledge, mostly around setting up test networks, a very tedious task that finally is quick to set up
* [TODO] allow 0 quorum record placements. as our cluster only has one node in the first place, this is necessary, acting more as shared-state across cluster nodes.
* [TODO] fix up the last failing test with IPFS in peer_to_peer.rs
https://github.com/subconsciousnetwork/noosphere/pull/290


jsantell:
This PR has been broken up and landed in several other pieces; closing

Issue #291, opened 2023-03-26
Integration test for `ns_sphere_traverse_by_petname`


cdata:
We punted on writing this because it needs a lot of extra work to get done. But, the work is worth doing, so let's get it done and write the test.

PR #292, opened 2023-03-26
feat!: Sphere traversal C FFI

cdata:
This change introduces a C FFI for sphere traversal.

This change also breaks a bunch of existing FFI by renaming functions to better align with the `SphereContext` refactor we did in #253. Only the names of functions have been changed. The semantics remain the same.

Additionally, the change fixes the C example, which has actually been broken for a little while.

I really wanted to include an integration test for this one, but it's actually a lot harder to do than it seems. We need to spin up two gateways and a name system node as separate processes in order to perform the test. This means that we would have to build both `orb` and `orb-ns` for the MacOS target before running the Swift tests and then write some helper code to spin them up during a test. In the end, we already cover the majority of the testable behavior in the equivalent Rust integration test. So, I've filed #291 for us to do the work down the road.

### DIFF
--- a/rust/noosphere/src/ffi/context.rs
+++ b/rust/noosphere/src/ffi/context.rs
@@ -83,7 +83,7 @@ impl NsSphereFile {
 /// This will fail if it is not possible to initialize a sphere with the given
 /// identity (which implies that no such sphere was ever created or joined on
 /// this device).
-pub fn ns_sphere_fs_open(
+pub fn ns_sphere_open(
     noosphere: &NsNoosphere,
     sphere_identity: char_p::Ref<'_>,
     error_out: Option<Out<'_, repr_c::Box<NsError>>>,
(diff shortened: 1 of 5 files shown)


rust/noosphere/src/ffi/context.rs
```diff
@@ -104,9 +104,47 @@ pub fn ns_sphere_fs_open(
     })
 }
 
+#[ffi_export]
+/// Access another sphere by a petname. The petname should be one that has been
```
jsantell, line 108: nit: add `\n\n` after the first sentence (e.g. having a whole row between the first sentence and the second sentence). This splits up the function documentation into "summary" and "detailed" information.

rust/noosphere/src/ffi/context.rs
```diff
@@ -175,9 +213,9 @@ pub fn ns_sphere_fs_read(
 /// slug, it will be assigned to be the previous historical revision of the new
 /// content.
 ///
-/// Note that you must invoke [ns_sphere_fs_save] to commit one or more writes
-/// to the sphere.
-pub fn ns_sphere_fs_write(
+/// Note that you must invoke [ns_sphere_save] to commit one or more writes to
```
jsantell, line 216: nit: add parentheses in function name to link doxygen `ns_sphere_save()`

rust/noosphere/src/ffi/context.rs
```diff
@@ -104,9 +104,47 @@ pub fn ns_sphere_fs_open(
     })
 }
 
+#[ffi_export]
+/// Access another sphere by a petname. The petname should be one that has been
+/// assigned to the sphere's identity using `ns_sphere_petname_set`. If any of
+/// the data required to access the target sphere is not available locally, it
+/// will be replicated from the network through a the configured Noosphere
+/// Gateway. If no such gateway is configured and the data is not available
+/// locally, this call will fail. The returned `NsSphere` pointer can be used to
```
jsantell, line 113: nit: use C types in FFI functions e.g. `ns_sphere_t`

jsantell (dismissed):
Great clean up of FFI type names. Some nits on docs in the context of generation #285

rust/noosphere/src/ffi/context.rs
```diff
@@ -104,9 +104,47 @@ pub fn ns_sphere_fs_open(
     })
 }
 
+#[ffi_export]
+/// Access another sphere by a petname. The petname should be one that has been
+/// assigned to the sphere's identity using `ns_sphere_petname_set`. If any of
```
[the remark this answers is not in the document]
cdata: That's a bit weird, since we'll be omitting the args, right?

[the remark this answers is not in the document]
jsantell: example: https://jsantell.dev/noosphere/classns__sphere__file__t.html#aaf5386f5e16ff15047956ad7d24ddc24

jsantell (approved):
:+1:

PR #296, opened 2023-03-31
fix: Clean up release workflows

cdata:
- Fix missing checksum in generated Package.swift
 - Fix custom container image tag not being tagged
 - Tag "latest" container images on every push to main
 - Tag "noosphere-$VERSION" container images on every release


jsantell (approved):
:raised_hands: nice!

PR #297, opened 2023-03-31
feat: Introduce `Link`, a typed `Cid`

cdata:
This new `Link` type is a re-interpretation of the same concept in `libipld`. Unfortunately, the `libipld` version has some qualities that discourage us from using it directly (it's not `serde`-friendly, and also has no awareness of block stores).

By using `Link` in place of `Cid`, we can have well-defined associations between data types in cases where the association is otherwise just implied (such as is the case with `SphereIpld`).

This change is a precursor to some additional changes where we will rename a bunch of fields and also partially revert breaking changes to our HAMT that were introduced in #284 .


jsantell (approved):
Excellent :tada:

PR #298, opened 2023-04-03
Last (hopefully) breaking changes ahead of beta

cdata:
This change implements the remaining breaking changes to the Noosphere core data structures that we hope to land ahead of the upcoming Private Beta phase of Subconscious. The breaking changes are several, and I'll attempt to document them here:

## Renames

The following core domain concepts have been renamed to be more coherent or consistent with the way we have come to talk about them among ourselves:

 - Links -> Content
 - Names -> AddressBook / Identities
 - Authorization -> Authority
 - Allowed UCANs -> Delegations
 - Revoked UCANs -> Revocations

## HAMT layout

The HAMT layout changes implemented in #284 have been almost entirely reverted. With the landing of the typed CID in #297 , we can now leverage the type system to be precise about what means when the key or the value of the HAMT is a CID. This revised approach does lead us to have to be more fiddle-y when crawling the sphere to stream its blocks, but otherwise it gives us improved data layout flexibility while maintaining the benefits of the change made in #284 .

## Other breaking changes

The following additional breaking changes were made:

 - More method names that were prefixed with `try_` have had that prefix removed to align with our convention elsewhere
 - Versions of the `fastcdc` and `base64` crates were updated by breaking increments


rust/noosphere-core/src/view/authority.rs
```diff
@@ -66,15 +66,15 @@ where
         })
     }
 
-    pub async fn try_get_allowed_ucans(&self) -> Result<AllowedUcans<S>> {
+    pub async fn get_delegations(&self) -> Result<Delegations<S>> {
```
jsantell, line 69: `delegations` much better than `allowed_ucans`, nice!

rust/noosphere-core/src/view/mutation.rs
```diff
@@ -47,20 +50,20 @@ impl<S: BlockStore> SphereRevision<S> {
...
 pub struct SphereMutation {
     did: String,
-    links: LinksMutation,
-    names: NamesMutation,
-    allowed_ucans: AllowedUcansMutation,
-    revoked_ucans: RevokedUcansMutation,
+    content: ContentMutation,
+    identities: IdentitiesMutation,
+    delegations: DelegationsMutation,
+    revocations: RevocationsMutation,
```
jsantell, line 56: Simple, and given the data spheres store, non-ambiguous, especially compared to "links"/"names" -- huge improvement in clarity here IMO

rust/noosphere-core/src/data/address.rs
```diff
@@ -1,91 +1,133 @@
...
-                            None => {
-                                warn!("No 'link' field in fact, skipping...");
+/// A [LinkRecord] is a newtype that represents a JWT that ought to contain a
+/// [Cid] reference to a sphere
+#[derive(Debug, Eq, PartialEq, Clone, Serialize, Deserialize, Hash)]
+#[serde(from = "Jwt", into = "Jwt")]
+#[repr(transparent)]
+pub struct LinkRecord(Jwt);
+
+impl LinkRecord {
```
jsantell, line 65: As a "core type", wonder if `noosphere_ns::NsRecord` (or renamed to `noosphere_ns::LinkRecord`) could be replaced by this, or sharing some functionality.

jsantell (approved):
Fantastic improvements and clarifications here @cdata!

rust/noosphere-core/src/data/address.rs
cdata: Yah, definitely possible!

PR #299, opened 2023-04-04
chore: Revise default logging

cdata:
This change disables the previously hard-coded default log filters that were kind of random, confusing and inappropriate for most cases. Now, a `RUST_LOG` environment variable filter is required in order to enable tracing-based logging.

Additionally, the tracing-related dependencies have been uplifted as "workspace" dependencies. My hope is that we can use this as a small first step towards uplifting ever more common dependencies to the workspace level.


jsantell (dismissed):
:+1: good to flip off the defaults, and see some workspace dependencies here too :partying_face:

jsantell (approved):

PR #300, opened 2023-04-05
fix: Introduce `TryOrReset` to help worker threads

cdata:
This introduces an `TryOrReset` construct that makes initialization of HTTP clients more resilient. In the case of both the name system and gateway APIs, client initialization is predicated on verifying the identity of the node we are talking to. `TryOrReset` enables us to express usage of a client that is resilient to failures related to identifying nodes, while mostly retaining the efficiency of the current pattern where client initialization only has to happen once.

### DIFF
--- a/rust/noosphere-gateway/src/again.rs
+++ b/rust/noosphere-gateway/src/again.rs
@@ -1,0 +1,147 @@
+use anyhow::Result;
+use std::future::Future;
+use std::sync::Arc;
+use tokio::sync::OnceCell;
+
+/// Wraps an "initialization" step, expressed as a closure, and allows
+/// a user to invoke closures with the result of that initialization step
+/// as a context argument. The result of invocation is always returned, but
+/// a failure result causes the initialized context to be reset is that it
+/// is re-initialized upon the next invocation attempt.
+pub struct Again<I, O, F>
+where
+    F: Future<Output = Result<O, anyhow::Error>>,
+    I: Fn() -> F,
+{
+    init: I,
+    initialized: OnceCell<Arc<O>>,
+}
+
+impl<I, O, F> Again<I, O, F>
+where
+    F: Future<Output = Result<O, anyhow::Error>>,
+    I: Fn() -> F,
+{
+    pub fn new(init: I) -> Self {
+        Again {
+            init,
+            initialized: OnceCell::new(),
+        }
+    }
+
+    /// Invoke a closure with the initialized context. The result will be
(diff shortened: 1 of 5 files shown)


rust/noosphere-gateway/src/again.rs
```diff
@@ -0,0 +1,147 @@
+use anyhow::Result;
+use std::future::Future;
+use std::sync::Arc;
+use tokio::sync::OnceCell;
+
+/// Wraps an "initialization" step, expressed as a closure, and allows
+/// a user to invoke closures with the result of that initialization step
+/// as a context argument. The result of invocation is always returned, but
+/// a failure result causes the initialized context to be reset is that it
+/// is re-initialized upon the next invocation attempt.
```
jsantell, line 10: Last sentence might be missing a word -- could be a good ASCII FSM diagram candidate

rust/noosphere-gateway/src/again.rs
```diff
@@ -0,0 +1,147 @@
...
+            init,
+            initialized: OnceCell::new(),
+        }
+    }
+
+    /// Invoke a closure with the initialized context. The result will be
+    /// returned as normal, but an error result will cause the initialized
+    /// context to be reset so that the next time an invocation is attempted,
+    /// context initialization will be retried.
+    pub async fn invoke_or_reset<Ii, Oo, Ff>(&mut self, invoke: Ii) -> Result<Oo>
```
jsantell, line 36: This function either invokes AND resets, or just invokes -- possibly rename to just `invoke`, as this whole struct/state is the "reset"

rust/noosphere-gateway/src/again.rs
```diff
@@ -0,0 +1,147 @@
...
+use std::future::Future;
+use std::sync::Arc;
+use tokio::sync::OnceCell;
+
+/// Wraps an "initialization" step, expressed as a closure, and allows
+/// a user to invoke closures with the result of that initialization step
+/// as a context argument. The result of invocation is always returned, but
+/// a failure result causes the initialized context to be reset is that it
+/// is re-initialized upon the next invocation attempt.
+pub struct Again<I, O, F>
```
jsantell, line 11: The name `Again` is a bit generic, evoking low level algebraic types or helpers. Brainstorming some other names, each with their own problems, but where my head's at:

* `SafeInitializer`
* `ContextInitializer`
* `ResetOnErr`
* `CleanContext`

jsantell (dismissed):
Fantastic, only naming suggestions from me :+1:

PR #302, opened 2023-04-05
Several updates for noosphere-ipfs

jsantell:
* IpfsStore::get_block now resolves blocks with the correct output-codec.
* IPFS tests (--features test_kubo) could produce false positives by relying on previously cached values in the IPFS instance. Test data now is randomized.
* Add real test against IPFS for a non-existant block via BlockStoreRetry and IpfsStore

### DIFF
--- a/rust/noosphere-ipfs/src/client/kubo.rs
+++ b/rust/noosphere-ipfs/src/client/kubo.rs
@@ -10,7 +10,20 @@ use hyper::{
 };
 use hyper_multipart_rfc7578::client::multipart::{Body as MultipartBody, Form};
 use ipfs_api_prelude::response::{IdResponse, PinLsResponse};
+use libipld_cbor::DagCborCodec;
+use libipld_core::raw::RawCodec;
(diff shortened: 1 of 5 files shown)


rust/noosphere-ipfs/src/client/kubo.rs
```diff
@@ -95,27 +108,30 @@ impl IpfsClient for KuboClient {
 
     async fn put_block(&mut self, _cid: &Cid, _block: &[u8]) -> Result<()> {
         unimplemented!();
+        /*
+        let mut car = Vec::new();
+        let mut car_writer = CarWriter::new(CarHeader::new_v1(vec![cid.to_owned()]), &mut car);
+        car_writer.write(cid.to_owned(), block).await?;
+        self.syndicate_blocks(Cursor::new(car)).await
+        */
```
jsantell, line 116: Was a utility for tests, can investigate further

cdata (approved):
Nice!

PR #303, opened 2023-04-06
feat: cache 'peer_id' in orb-ns to provide a HTTP route that does not lock the NS mutex for testing.

jsantell:


cdata (approved):
Nice, looks good!

PR #304, opened 2023-04-06
feat: Add additional tracing instrumentation to noosphere-ns and noosphere-ipfs

jsantell:
* Adds tracing to relevant DHT commands, IPFS lookups, and record validations. Future improvements could include spans that follow DHT operations logically (responses from the DHT are polled and reconciled with one or many top-level requests), and better formatting of Cids/Records.
* Remove some redundant, ill-formatted tracing


cdata (approved):
Awesome :beers:

PR #310, opened 2023-04-11
fix: Unreachable petname sequence is not an error

cdata:
This change makes double-darn sure that we aren't panicking unexpectedly when dealing with slashlinks in the FFI. It also changes some of the error semantics of traversal. Previously, trying to traverse to a petname that isn't assigned to a DID would result in an error. Now, it results in a `None` (similar to how reading a slug that isn't assigned results in `None`).


cdata:
Will defer landing this until #313 lands

jsantell (approved):
LGTM

PR #313, opened 2023-04-12
feat: initial attempt at connecting C tests to CI

jsantell:
Add C tests to CI, and fix interpreting uint8 data as char*, as well as a leak.

### DIFF
--- a/.github/workflows/run_test_suite.yaml
+++ b/.github/workflows/run_test_suite.yaml
@@ -82,7 +82,26 @@ jobs:
         with:
           command: test
           args: --features test_kubo,headers
-
+  run-test-suite-linux-c:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: Swatinem/rust-cache@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          override: true
+          profile: minimal
+          toolchain: stable
+      - name: 'Install environment packages'
+        run: |
+          sudo apt-get update -qqy
+          sudo apt-get install jq protobuf-compiler cmake make
+      - name: 'Link Noosphere from C'
+        run: |
+          make build -C ./c/example
+      - name: 'Run C integration tests'
+        run: |
+          make run -C ./c/example
   run-test-suite-web-wasm:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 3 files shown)


jsantell:
The build process pulls `libnoosphere.a` from cache, though requires `noosphere.h` generation, which compiles the related example.

.github/workflows/run_test_suite.yaml
```diff
@@ -82,7 +82,10 @@ jobs:
         with:
           command: test
           args: --features test_kubo,headers
-
+      - name: 'Run C tests'
```
cdata, line 85: Can we make this it's own "job" (as opposed to a "step" within a "job")? This way it will show up as its own line in the PR status check area.

cdata (changes requested):
Looks good, minor stylistic change requested

.github/workflows/run_test_suite.yaml
jsantell: yeah!

jsantell: Haven't yet looked into caching yet, but a separate job results in a whole new env install and recompile; maybe rebranded as "FFI integration tests" would be more appropriate as a step within `run-test-suite-linux`?

cdata: I think it's good for us to split all the distinctive test runs out. Eventually I would like us to be able to run them as stand-alone workflows.

.github/workflows/run_test_suite.yaml
```diff
@@ -82,7 +82,24 @@ jobs:
...
+      - uses: actions-rs/toolchain@v1
+        with:
+          override: true
+          profile: minimal
+          toolchain: stable
+      - name: 'Install environment packages'
+        run: |
+          sudo apt-get update -qqy
+          sudo apt-get install jq protobuf-compiler cmake
+      - name: 'Run C tests'
```
cdata, line 99: Can we also build the example as a step in the test run? This will make sure our example doesn't fall out of date (kind of like a doc test of sorts).

cdata (changes requested):

.github/workflows/run_test_suite.yaml
jsantell: In `run-test-suite-linux`? Ya, but then I don't understand what this job is providing (and could just be thought of as integration tests for the rust FFI layer, rather than a C environment), to just be a part of e.g. `run-test-suite-linux`. As of now the example IS the test

.github/workflows/run_test_suite.yaml
jsantell: alternatively, these could also live as `rust/noosphere/tests/*.c` bootstrapped by cargo test, testing the rust library moreso than a bindings environment

.github/workflows/run_test_suite.yaml
cdata: No, I want you to do it here as a step before or after the test run.

cdata: I assume that the example has the same binary dependencies as the tests.

jsantell: Ah I gotcha now, split up the build from the run in their own steps, sg!

cdata (approved):
Nice!

PR #316, opened 2023-04-12
fix: Intermittent timeouts in DhtNode tests introduced via #308 by adding peers before listening.

### LINKED ISSUE #311: [Intermittent] dht::node::test::test_dhtnode_bootstrap
This occurs occasionally, has happened at least for both Linux and Windows builds

```
---- dht::node::test::test_dhtnode_bootstrap stdout ----
thread 'dht::node::test::test_dhtnode_bootstrap' panicked at 'timed out: waiting for 4 peers', rust\noosphere-ns\src\dht\node.rs:243:42
```

jsantell:
Was able to reproduce this locally after many tests runs. Was not able to repro locally after this after many test runs. Fixes #311

### DIFF
--- a/rust/noosphere-ns/src/dht/node.rs
+++ b/rust/noosphere-ns/src/dht/node.rs
@@ -257,6 +257,8 @@ mod test {
         try_join_all(futures).await
     }
 
+    /// Creates a network of `node_count` nodes, with all nodes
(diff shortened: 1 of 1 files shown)


cdata (approved):
Thanks!

PR #317, opened 2023-04-14
Introduce `ns_sphere_identity` FFI call

bfollington:
Context:

>  Right now when I call `ns_sphere_traverse_by_petname` I get back a pointer to a `ns_sphere_t` which we wrap in our own `Sphere` Swift type. That is all good except that `Sphere` needs to know the DID of the `ns_sphere_t` it holds, but (afaict?) there is no easy way for me to get it reliably when traversing to a fully qualified petname (e.g. `@ben.chris.jordan`).

> I find myself wishing for either:
> 1. `ns_sphere_identity_get` so I can read the DID out of the sphere post-traversal
> 2. `ns_resolve_did` which can translate from a fully qualified petname to a DID (resolve is overloaded as a term)

> Leaning towards (1). I can work around this by traversing one hop at a time in a loop and retrieving each DID along the way using `ns_sphere_petname_get` but it seems a shame when traversing can handle dotted petnames.

# Changes

- Introduce `ns_sphere_identity` FFI export

### DIFF
--- a/rust/noosphere/src/ffi/context.rs
+++ b/rust/noosphere/src/ffi/context.rs
@@ -530,7 +530,7 @@ pub fn ns_sphere_file_version_get(
 }
 
 #[ffi_export]
-/// @memberof ns_sphere_file_t
+/// @memberof ns_sphere_t
 /// Get the identity (a DID encoded as a UTF-8 string)
 /// for this ns_sphere_t.
 pub fn ns_sphere_identity(
--- a/rust/noosphere/src/ffi/context.rs
+++ b/rust/noosphere/src/ffi/context.rs
@@ -528,3 +528,21 @@ pub fn ns_sphere_file_version_get(
             .map_err(|error: InvalidNulTerminator<String>| anyhow!(error).into())
     })
 }
+
+#[ffi_export]
+/// @memberof ns_sphere_file_t
+/// Get the identity (a DID encoded as a UTF-8 string)
+/// for this ns_sphere_file_t.
+pub fn ns_sphere_file_identity_get(
+    sphere_file: &NsSphereFile,
+    error_out: Option<Out<'_, repr_c::Box<NsError>>>,
+) -> Option<char_p::Box> {
+    error_out.try_or_initialize(|| {
+        sphere_file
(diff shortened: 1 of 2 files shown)


cdata:
Looks good, perhaps we could add a test to https://github.com/subconsciousnetwork/noosphere/blob/main/swift/Tests/SwiftNoosphereTests/NoosphereTests.swift?

rust/noosphere/src/ffi/context.rs
```diff
@@ -528,3 +528,28 @@ pub fn ns_sphere_file_version_get(
             .map_err(|error: InvalidNulTerminator<String>| anyhow!(error).into())
     })
 }
+
+#[ffi_export]
+/// @memberof ns_sphere_file_t
+/// Get the identity (a DID encoded as a UTF-8 string)
+/// for this ns_sphere_t.
+pub fn ns_sphere_identity_get(
```
jsantell, line 536: `ns_sphere_identity()` may be more consistently named (e.g. `ns_sphere_receipt_identity()` as similar on a `ns_sphere_receipt_t` object)

cdata:
Second @jsantell 's naming feedback, otherwise looks great!

rust/noosphere/src/ffi/context.rs
bfollington: Done!

rust/noosphere/src/ffi/context.rs
```diff
@@ -528,3 +528,28 @@ pub fn ns_sphere_file_version_get(
             .map_err(|error: InvalidNulTerminator<String>| anyhow!(error).into())
     })
 }
+
+#[ffi_export]
+/// @memberof ns_sphere_file_t
```
cdata, line 533: Sorry, almost missed this: should be changed to `ns_sphere_t`

bfollington: Nice catch!

cdata (approved):
Yay, thanks for this contribution!

cdata (approved):
So nice I approved it twice

PR #326, opened 2023-04-22
Update dependencies

### LINKED ISSUE #208: Update to `libipld` v0.16.x
It's that time again. `libipld` has released another major revision, which means the ecosystem we depend on will probably be updating shortly. The main change is described here: https://github.com/ipld/libipld/pull/174

We need to update more-or-less in step with other crates like `wnfs` that also depend on this crate.

matheus23:
I wanted to update `noosphere-car` so it works with `libipld` v0.16.
This sent me on a path to upgrading a bunch of things :sweat_smile: 

This PR depends on https://github.com/ucan-wg/rs-ucan/pull/78

Also: I had to update `libp2p`, because of errors with incompatible `digest` bounds. Then I was getting weird type errors within the `libipld-connection-limits` package, which I could fix by deleting `Cargo.lock` and re-creating it, but it's quite a large hammer.

Not sure how to best test all of this, but wanted to throw this PR out there since I think it helps a bunch already :)

Closes #208


cdata:
Thanks for taking this on @matheus23 the CI automation should cover libp2p a bit. I'll test it locally as well.

cdata:
I'm going to branch your change and do some cleanup on it now that your rs-ucan patch is released in `ucan = "0.1.2"`.

jsantell:
> Also: I had to update `libp2p`, because of errors with incompatible `digest` bounds. Then I was getting weird type errors within the `libipld-connection-limits` package, which I could fix by deleting `Cargo.lock` and re-creating it, but it's quite a large hammer.

Was just looking at updating libp2p to 0.51.3 and there were a handful of issues I've noted below -- if that's a prereq for this, I can take a look Monday. Clobbering cargo lock and auditing the diff probably the way to go.

>  * Upgrading libp2p to 0.51.3 -- PeerId and Keypair collisions from underlying libp2p-core 0.39.0 (though 0.39.1 is published with newer referenced things.)
 >  * Version uses dep libp2p-core 0.39.0, though seems to require features in 0.39.1 (published), which reconciles two impls of PeerId; currently, as published on Cargo, not sure how this can work, with errors (in both directions) inside newly created internal libp2p crates: expected struct libp2p_core::PeerId, found struct libp2p_identity::PeerId.
 > * Keypair errors squashed by mostly remapping to core::identity::Keypair, though may be some struggles incoming

cdata:
@jsantell there were a few warnings related to the libp2p change, but since everything builds and tests pass it seemed prudent to land this; otherwise we could not update the other IPLD dependencies.

If you are aware of any sharp corners in the v0.51.3 updated, we can get a fix in for them quickly on Monday.

cdata:
Going to close this since the work landed in #327

PR #329, opened 2023-04-24
chore: Fix deprecation notices from libp2p 0.51.3

jsantell:
* Consistently use `libp2p::identity::Keypair` rather than `libp2p::core::identity::Keypair` and use the `libp2p-identity` crate recently introduced. Rework some of the decoding methods.
  * New keys use secret key only for de/encoding -- the `ed25519_to_bytes` helper could be configured to use this, or removed completely if inlined into the `NameSystemKeyMaterial` trait.
* Remove deprecated/unused mplex module
* Replace `Swarm::with_tokio_executor`
* New `libp2p_allow_block_list` module requires more configuring with our mixed `DhtBehavior` (implementing for both kad/identify underlying behaviors), so `SwarmEvent::BannedPeer` event remains deprecated

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2874,7 +2874,6 @@ dependencies = [
  "libp2p-kad",
  "libp2p-mdns",
  "libp2p-metrics",
- "libp2p-mplex",
  "libp2p-noise",
  "libp2p-quic",
  "libp2p-swarm",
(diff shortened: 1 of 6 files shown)


cdata (approved):
Thanks for this cleanup!

PR #334, opened 2023-04-26
Store name records in gateway for repropagation. Fixes #330

### LINKED ISSUE #330: Persist Records between gateway and NS
The NS built on libp2p replicates and publishes records with a reasonable TTL, essentially wrapping our JWT records to be available over the distributed network. (More details https://github.com/subconsciousnetwork/noosphere/issues/309).
libp2p uses their [RecordStore trait](https://docs.rs/libp2p/latest/libp2p/kad/record/store/trait.RecordStore.html) to maintain these records, giving us an opportunity to persist records between NS restarts.

### Suggested solution

Store LATEST_NAME_RECORD for each managed sphere in the gateway's counterpart sphere. Gateway storage persists already. On gateway startup, inform NS which records it should provide. This results in the NS being "stateless", and kicks more responsibility to the gateway, but seems like a reasonable amount for the gateway to do between an ephemeral NS, and a local sphere that cannot store its own records.

Continue using `libp2p::kad::record::store::MemoryStore` within libp2p to manage records (this includes both "provided" records, and records from other nodes). We could implement `RecordStore`, backed by a Noosphere Storage type (like `NativeStorage`), though the noosphere types are all async, while libp2p's storage is sync and could blockingly wrap e.g. `NativeStorage` within the libp2p interface. Though Expiry and replication is already handled here for both provided/other records, and quite complex. Losing these intermediate records upon process restart is OK, and even desired as network changes (e.g. in complex topologies, using yesterday's records may be actively harmful).

Some initial work done here: https://github.com/subconsciousnetwork/noosphere/compare/publish-record-on-gateway-start

jsantell:


rust/noosphere-cli/tests/peer_to_peer.rs
```diff
@@ -222,31 +250,40 @@ async fn gateway_publishes_and_resolves_petnames_configured_by_the_client() {
         third_party_gateway_task.abort();
     });
     client_task.await.unwrap();
+
+    // Restart gateway and name system, ensuring republishing occurs
+    let (ns_task, ns_url) = start_name_system_server(&ipfs_url).await.unwrap();
```
cdata, line 255: Clean up warnings here

See what we do here: https://github.com/subconsciousnetwork/noosphere/blob/6c6b8d92db859ccd1dc8bd6055671b7db82596e3/rust/noosphere-cli/tests/peer_to_peer.rs#L248-L250

rust/noosphere-cli/tests/peer_to_peer.rs
```diff
@@ -222,31 +250,40 @@ async fn gateway_publishes_and_resolves_petnames_configured_by_the_client() {
...
+    let ns_client = HttpClient::new(ns_url.clone()).await.unwrap();
+    assert!(
+        ns_client.resolve(&client_identity).await.unwrap().is_none(),
+        "new name system does not contain client identity"
+    );
+
+    let (gateway_url, gateway_task) = start_gateway_for_workspace(
```
cdata, line 262: Clean up warnings here

rust/noosphere-gateway/src/worker/name_system.rs
```diff
@@ -103,6 +114,53 @@ where
...
+    K: KeyMaterial + Clone + 'static,
+    S: Storage + 'static,
+{
+    loop {
+        for local_sphere in local_spheres.iter() {
+            if let Err(error) = periodic_publish_record(&tx, local_sphere).await {
+                error!("Could not publish record: {}", error);
+            };
+        }
+        tokio::time::sleep(Duration::from_secs(PERIODIC_PUBLISH_INTERVAL_SECONDS)).await;
```
cdata, line 134: Can you also pull the time for the other periodic task out into its own const so that we can see them together at a glance?

rust/noosphere-gateway/src/worker/name_system.rs
```diff
@@ -376,3 +437,60 @@ impl<H> OnDemandNameResolver<H> {
...
+    };
+    let counterpart_link_record_key = format!("link_record/{counterpart_identity}");
+    let mut cursor = SphereCursor::latest(context.clone());
+    let (_, bytes) = block_serialize::<DagCborCodec, _>(record)?;
+
+    cursor
+        .write(
+            &counterpart_link_record_key,
+            &ContentType::Cbor.to_string(),
+            Cursor::new(bytes),
```
cdata, line 461: TBH if you are just storing the JWT, the CBOR encoding is kind of overkill. Might as well just store UTF-8 string bytes :shrug: 

Encoding as CBOR is nice for structured data. But, `Jwt` is just a wrapper around a `String`. So...

```rs
cursor.write(
    slug,
    content_type,
    record.as_bytes(),
    None,
).await?;
```

...should suffice. Then, when reading:

```rs
let mut buffer = String::new();
let mut file = context.read(...).await?.unwrap_or_else(...)?;
file.context.read_to_string(&mut buffer).await?;
let jwt = Jwt(buffer);
```

cdata (changes requested):
Looks good, some improvements recommended!

rust/noosphere-gateway/src/worker/name_system.rs
cdata: If we implemented `DerefMut` for our string wrappers we could even pass a mutable `Jwt` instead of a naked string when reading.... :thinking:

rust/noosphere-gateway/src/worker/name_system.rs
```diff
@@ -376,3 +436,53 @@ impl<H> OnDemandNameResolver<H> {
...
+        let sphere_context = context.sphere_context().await?;
+        let db = sphere_context.db();
+        db.require_key::<_, Did>(COUNTERPART).await?
+    };
+    let counterpart_link_record_key = format!("link_record/{counterpart_identity}");
+    let mut cursor = SphereCursor::latest(context.clone());
+    cursor
+        .write(
+            &counterpart_link_record_key,
+            &ContentType::Bytes.to_string(),
```
cdata, line 457: Probably should make this "text/plain" (which I'm just now noticing we don't have a `ContentType` variant for :sweat: ).

cdata (changes requested):

cdata (approved):
Woot!

PR #335, opened 2023-04-26
feat: Enable support of quorum=0 for DHT during infra bootstrapping phase

### LINKED ISSUE #264: Name system HTTP client ignores server JSON response when putting records
Currently the HTTP client for the name system operator's API ignores the server response when putting a new record. The main reason for this is that the trait it implements does not call for any return type from the associated client method (just the unit type). We should evaluate if this is correct and determine what the expected return value should be.

jsantell:
* Decouple validity vs expiration in NsRecord::validate.
* Fixes #264, using response from name system server.
* Remove "local cache" from NameSystem, as its a bit dated and redundant with libp2p Kademlia storage, where persistence and validity should be handled.
* Rename s/NameSystemClient/DhtClient and separate p2p nature from the name resolver, via NameResolver trait (publish, resolve) for consumer usage (e.g. hiding those quorum rules from gateway).

### DIFF
--- a/rust/noosphere-ns/src/name_system.rs
+++ b/rust/noosphere-ns/src/name_system.rs
@@ -2,15 +2,12 @@ use crate::{
     dht::{DhtConfig, DhtError, DhtNode, DhtRecord, NetworkInfo, Peer},
     records::{NsRecord, RecordValidator},
     utils::make_p2p_address,
-    NameSystemClient, PeerId,
+    DhtClient, PeerId,
 };
 use anyhow::{anyhow, Result};
 use async_trait::async_trait;
-use futures::future::try_join_all;
 use libp2p::{identity::Keypair, Multiaddr};
 use noosphere_core::data::Did;
-use std::collections::HashMap;
-use tokio::sync::{Mutex, MutexGuard};
 use ucan::{crypto::KeyMaterial, store::UcanJwtStore};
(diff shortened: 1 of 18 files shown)


rust/noosphere-ns/src/name_system.rs
```diff
@@ -2,15 +2,12 @@ use crate::{
     dht::{DhtConfig, DhtError, DhtNode, DhtRecord, NetworkInfo, Peer},
     records::{NsRecord, RecordValidator},
     utils::make_p2p_address,
-    NameSystemClient, PeerId,
+    DhtClient, NameResolver, PeerId,
```
cdata, line 5: Fix warning

rust/noosphere-ns/tests/ns_test.rs
```diff
@@ -7,7 +7,7 @@ use noosphere_core::{authority::generate_ed25519_key, data::Did, view::SPHERE_LI
 use noosphere_ns::{
     helpers::NameSystemNetwork,
     utils::{generate_capability, generate_fact, wait_for_peers},
-    NameSystemClient, NsRecord,
```
cdata, line 10: Fix warnings throughout

cdata (changes requested):
Looks good, just needs a little cleanup

jsantell:
Rebased and lints fixed! Going through the tests once more

cdata (approved):

PR #338, opened 2023-04-27
feat: validate link records on the gateway

### LINKED ISSUE #257: Verify proof chain for name record adoption candidates

jsantell:
Enables gateways to validate a records' proof chain before adoption over IPFS, and verify expiry before publishing. Fixes #257. Changes in `syndication.rs` are exclusively from moving the logic inside a function to catch errors without taking down the processor (only logic change is early aborting if IPFS node isn't found).

### DIFF
--- a/rust/noosphere-gateway/src/worker/name_system.rs
+++ b/rust/noosphere-gateway/src/worker/name_system.rs
@@ -5,17 +5,15 @@ use cid::Cid;
 use noosphere_core::data::ContentType;
 use noosphere_core::data::{Did, IdentityIpld, Jwt, LinkRecord, MapOperation};
 use noosphere_ipfs::{IpfsStore, KuboClient};
-use noosphere_ns::NsRecord;
-use noosphere_ns::{server::HttpClient as NameSystemHttpClient, NameResolver};
-use noosphere_sphere::SphereContentRead;
-use noosphere_sphere::SphereContentWrite;
-use noosphere_sphere::COUNTERPART;
+use noosphere_ns::{server::HttpClient as NameSystemHttpClient, NameResolver, NsRecord};
 use noosphere_sphere::{
     HasMutableSphereContext, SphereCursor, SpherePetnameRead, SpherePetnameWrite,
 };
+use noosphere_sphere::{SphereContentRead, SphereContentWrite, COUNTERPART};
 use noosphere_storage::KeyValueStore;
 use noosphere_storage::{BlockStoreRetry, Storage, UcanStore};
 use std::fmt::Display;
+use std::future::Future;
 use std::{
     collections::{BTreeMap, BTreeSet, VecDeque},
     str::FromStr,
@@ -86,7 +84,11 @@ pub enum NameSystemJob<C> {
     /// revision
     ResolveSince { context: C, since: Option<Cid> },
     /// Publish a link record (given as a [Jwt]) to the name system
-    Publish { context: C, record: Jwt },
+    Publish {
+        context: C,
+        record: Jwt,
+        temporary_validate_expiry: bool,
+    },
 }
 
 pub fn start_name_system<C, K, S>(
@@ -151,6 +153,7 @@ where
             if let Err(error) = tx.send(NameSystemJob::Publish {
                 context: local_sphere.to_owned(),
                 record,
+                temporary_validate_expiry: false,
             }) {
                 warn!("Failed to request name record publish: {}", error);
             }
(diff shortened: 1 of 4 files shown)


rust/noosphere-gateway/src/worker/syndication.rs
```diff
@@ -77,162 +80,178 @@ where
 
     let kubo_client = Arc::new(KuboClient::new(&ipfs_api)?);
 
-    while let Some(SyndicationJob { revision, context }) = receiver.recv().await {
-        debug!("Attempting to syndicate version DAG {revision} to IPFS");
-        let kubo_identity = match kubo_client.server_identity().await {
-            Ok(id) => id,
+    while let Some(job) = receiver.recv().await {
+        match process_task(job, kubo_client.clone(), &ipfs_api).await {
```
cdata, line 84: Probably should work "job" into the name here, like `process_job`.

rust/noosphere-gateway/src/worker/name_system.rs
```diff
@@ -150,7 +150,12 @@ where
             debug!("Running {}", job);
             match job {
                 NameSystemJob::Publish { record, .. } => {
-                    client.publish(NsRecord::from_str(&record)?).await?;
+                    let link_record = NsRecord::from_str(&record)?;
```
cdata, line 153: What are the chances we could get tests that cover the valid / invalid cases?

cdata (changes requested):
Mostly just name bike shedding, looks good!

rust/noosphere-gateway/src/worker/syndication.rs
```diff
@@ -61,10 +61,13 @@ where
...
 
-    (tx, tokio::task::spawn(ipfs_syndication_task(ipfs_api, rx)))
+    (
+        tx,
+        tokio::task::spawn(ipfs_syndication_processor(ipfs_api, rx)),
+    )
 }
 
-async fn ipfs_syndication_task<C, K, S>(
+async fn ipfs_syndication_processor<C, K, S>(
```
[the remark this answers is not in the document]
jsantell: SG! I'll change back to **task** a la Tokio, processing **jobs**

jsantell:
Dependent on https://github.com/ucan-wg/rs-ucan/pull/83

jsantell:
* Depends on #360 (ucan 0.2.0, validation that doesn't include expiry)
* Introduced `KeyValueNameResolver`, a simplified name resolver for unit tests
* Add a test for publishable timeframe for gateway, also illustrating how we can test individual job processing

rust/noosphere-gateway/src/worker/syndication.rs
```diff
@@ -77,162 +77,175 @@ where
```
jsantell, line 77: No logic changes in this file, just copies the handling of `process_job()` like the NS worker

rust/noosphere-gateway/src/worker/name_system.rs
```diff
@@ -147,115 +147,132 @@ where
...
+    I: Fn() -> F,
+    O: NameResolver + 'static,
+    F: Future<Output = Result<O, anyhow::Error>>,
+{
+    let run_job = with_client.invoke(|client| async move {
+        debug!("Running {}", job);
+        match job {
+            NameSystemJob::Publish { record, .. } => {
+                let link_record = NsRecord::from_str(&record)?;
+                if link_record.has_publishable_timeframe() {
```
jsantell, line 177: This file is noisy due to restructuring processing: adding a publishable timeframe check here

rust/noosphere-gateway/src/worker/name_system.rs
```diff
@@ -292,16 +307,12 @@ where
         let next_record =
             match fetch_record(client.clone(), name.clone(), identity.did.clone()).await? {
                 Some(record) => {
```
jsantell, line 309: This file is noisy due to restructuring processing: adding validation check here

rust/noosphere-cli/tests/peer_to_peer.rs
```diff
@@ -8,6 +8,8 @@ use noosphere_ipfs::{IpfsStore, KuboClient};
 use noosphere_storage::{BlockStoreRetry, MemoryStore, UcanStore};
```
jsantell, line 8: Changes here are shared with #344

jsantell:
Put validation behind a flag for deployment so the validation doesn't yet actually occur, but can be reconciled with other NS work and enabled shortly in the future

cdata:
This looks good, but I just want to clarify that it doesn't fix #257 . In order to fix #257, we need to introduce validation here: https://github.com/subconsciousnetwork/noosphere/blob/14f7d61209ae975928cc3c24f1df1dc55a4427ff/rust/noosphere-sphere/src/petname/write.rs#L93-L94

Sorry, we should have had those TODOs annotated.

This addresses perhaps one half of the task, but really validation should be happening at this point where we try to write to the SphereContext, and it should probably be mostly symmetrical across any sphere mutation (be it the gateway's sphere, the user's sphere, or ....)

cdata (approved):
LGTM

PR #340, opened 2023-04-28
feat: add ns_error_code to FFI. Fixes #332

### LINKED ISSUE #332: NoosphereError enum variants should be able to be one-way converted to int
In order to help FFI consumers (and others) take action on errors, we should enable errors to be one-way converted to unique error codes. This will be a lossy conversion (as errors can contain context-specific information that will not be trivially encoded in an int), but it will at least give FFI consumers a coarse category to branch on when reacting to errors.

jsantell:


rust/noosphere/src/error.rs
```diff
@@ -11,9 +14,19 @@ pub enum NoosphereError {
 
     #[error("Missing configuration: {0}")]
     MissingConfiguration(&'static str),
+}
 
-    #[error("{0}")]
-    Other(anyhow::Error),
+impl NoosphereError {
+    /// An error code representing the error. These mappings should never
+    /// change between versions and should be append only!
```
cdata, line 21: I don't think this is the correct contract to set with the caller. All API is subject to change between versions, including which numbers map to which error codes. If the change is a breaking one (e.g., we change an error code), the next release will have the appropriate Semver increment.

It would be capricious to break the API contract without good reason, so the chance of these things changing is relatively low. But, we definitely reserve the right to change these codes and their meaning over time, as called for by future research, development and discovery.

cdata (changes requested):
Looks good, thanks for taking this on. Some updates to the docs requested.

rust/noosphere/src/error.rs
cdata: One possible remedy to instability in these codes: export their values as constants. That way, a caller can compare against a named constant that can update over time as these codes update.

jsantell: Relaxing the ability to not change sounds good, and same with exposing constants :+1:

cdata (dismissed):
Awesome :raised_hands:

jsantell:
Updated using enums (safer_ffi does not support `#define`s or constant exports), and moved implementation to FFI (NsError) to avoid 1) circular dependencies and 2) redundant declarations in both NoosphereError and NsError where we could potentially miss exporting an error type. Not thrilled with the naming constraints on the export, but avoids conflicts with `ns_error`, `ns_error_code` etc.

jsantell:
Changes to FFI:
* Renaming `char* ns_error_string(const ns_error_t*)` to `char* ns_error_message_get(const ns_error_t*)`
* New method `uint8_t ns_error_code_get(const ns_error_t*)`

Swift example:
```swift
        let error_code = ns_error_code_get(maybe_error.pointee)
        let error_message_ptr = ns_error_message_get(maybe_error.pointee)
        let error_message = String.init(cString: error_message_ptr!)

        assert(error_code == NS_ERROR_CODE_OTHER.rawValue)
  ```

**Comment by jsantell** [comment]
> @jsantell after trying out different patterns for tracing to see how ergonomic they were on the Swift side, I aligned on a good one where Swift users don't have to deal in magic numbers and on the Rust side we can convert ints to enums before dealing with them.
> 
> Would you be willing to update this change to reflect that pattern?
> 
> https://github.com/subconsciousnetwork/noosphere/pull/342/files#diff-209a32115ba134ccb66539da851b87cf7a203ea55536385503d42698151a2785R16-R59

Updated based off of the style in the logging -- slightly different ergonomics here (enums OUT rather than enums IN), but similar. Similarly if the logging enums are `u32` for Swift ergo, we can update the enums here as well

File: rust/noosphere/src/ffi/error.rs [hunk 2]
```diff
@@ -1,13 +1,86 @@
...
+const NOOSPHERE_ERROR_OTHER: u8 = 1;
+const NOOSPHERE_ERROR_NETWORK_OFFLINE: u8 = 2;
+const NOOSPHERE_ERROR_NO_CREDENTIALS: u8 = 3;
+const NOOSPHERE_ERROR_MISSING_CONFIGURATION: u8 = 4;
+
+#[ffi_export]
+#[derive_ReprC(rename = "ns_error_code")]
+#[repr(u8)]
+/// Constant values for error codes from ns_error_t.
+pub enum NsErrorCode {
```
cdata, line 19: Given that the representation on the Swift side *must* be `u32`, let's change all the representations to reflect that. Otherwise it's a bit confusing.

cdata (changes requested):

cdata (approved):
Woot!

PR #341, opened 2023-04-28
chore: Add cargo fmt/clippy tests in Linux CI.

jsantell:


rust/noosphere-core/src/data/link.rs
```diff
@@ -67,6 +67,17 @@ where
     }
 }
 
+impl<T> PartialEq for Link<T>
```
jsantell, line 70: https://www.philipdaniels.com/blog/2019/rust-equality-and-ordering/

rust/noosphere-collections/src/hamt/node.rs
```diff
@@ -2,6 +2,7 @@
 // Source copyright and license:
 // Copyright 2019-2022 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
+#![allow(clippy::only_used_in_recursion)]
```
jsantell, line 5: There's a warning w/r/t `depth` argument in the below functions which isn't currently used. We could remove it, though there may be a future need for it, deferring to you @cdata

.github/workflows/run_test_suite.yaml
```diff
@@ -67,11 +67,24 @@ jobs:
         with:
           override: true
           profile: minimal
+          components: rustfmt, clippy
           toolchain: stable
       - name: 'Install environment packages'
         run: |
           sudo apt-get update -qqy
           sudo apt-get install jq protobuf-compiler cmake
+      - name: 'Check Format'
```
cdata, line 76: Let's run this as a separate job. That way it will get its own status line, so we can see if both A) lints pass and B) tests pass.

The jobs can run in parallel anyway.

cdata (changes requested):
Nice! This is long overdue. Just a minor tweak requested.

cdata (approved):
Awesome!

Issue #320, opened 2023-04-14
"No version has been resolved" when traversing multiple spheres


bfollington:
Attempting to traverse to any address that features a dotted petname path fails to resolve a version, even when that version should be known.

# Simple Reproduction

1. Create sphere `alice` 
2. Create sphere `bob` 
3. Create sphere `charlie`
4. `alice` follows `bob`
5. `bob` follows `charlie`
6. `alice` **can** traverse to `bob`
7. `bob` **can** traverse to `charlie`
8. `alice` **cannot** traverse to `charlie.bob`

# Expected Behaviour

`alice` _can_ traverse to `charlie.bob` since `bob` has resolved a CID for `charlie` already.

# Original Context

In our test environment I replicated this setup between `ziggy`, `flubbo` and `cdata-iphone`. When I traverse from `ziggy` to `flubbo` and then traverse on to `cdata-iphone` (so `@cdata-iphone.flubbo` from `ziggy`'s perspective) I get hit with:
 "**Failed to fetch profile: Foreign Error: No version has been resolved for "cdata-iphone" (`did:key:z6MkhA4679EbGpVCrWHpjcAg5V6gWxziA4d54FzgB8qDypSc`)**"

I'm confused by this, since if I boot up as `flubbo` they can traverse to `cdata-iphone` so they *must* have resolved a version? The same thing happens from any path with more than one step from the base sphere for me. I've made sure to make changes and sync everyone with their gateway (I believe).

PR #342, opened 2023-04-29
feat!: Revised tracing configuration

cdata:
This change overhauls how tracing configuration works for non-Wasm targets. A new C FFI has been added so that tracing can be initialized from embedders of the library. Also, most usages of `println` have been replaced with `tracing::info` as of this change.

The new semantics are reflected best by the comment docs for `initialize_tracing`:

```markdown
Initialize tracing-based logging throughout the Noosphere body of code,
as well as dependencies that implement tracing-based logging.

Invoking this function causes logs to be rendered until the termination
of the program. The default behavior is for logs to be rendered to
stdout. If this function is never called, logs will never be rendered.
Invoking this function more than once has no effect.

The function accepts an optional [NoosphereLog] preset configuration
that controls log filter level and display format.

Although this function accepts an optional preset configuration, it is
also sensitive to specific environment variables. The following
environment variables may be used to configure log behavior:

 - **`RUST_LOG`**: A comma-separated list of directives as described in
   [`tracing-subscriber`][1] and [`env-logger`][2] documentation
 - **`NOOSPHERE_LOG`**: An optional preset value interpretted as a
   [NoosphereLog]
 - **`NOOSPHERE_LOG_LEVEL`**: A specific filter level that, if set, is
   assigned to all of the [NOOSPHERE_LOG_LEVEL_CRATES]
 - **`NOOSPHERE_LOG_FORMAT`**: A specific format that, if set, is
   interpretted as a [NoosphereLogFormat] and configured as the format
   for log output

The configuration semantics are intended to be backwards compatible with
those defined by `tracing-subscriber` and `env-logger`. If only
`RUST_LOG` is set, then that configuration is applied as normal. The
[NoosphereLogFormat::Verbose] format is used in this case.

If some preset is given in the function invocation, that preset is
decomposed to determine how to further modify the log filter level and
format. The directives will be added to the ones prescribed by
`RUST_LOG`. If a `NOOSPHERE_LOG` environment variable is specified, it
will take precedence over preset given as an argument.

Finally, if a `NOOSPHERE_LOG_LEVEL` and/or `NOOSPHERE_LOG_FORMAT`
environment variable is configured, they will take final precedence to
determine the log filter level and format respectively.

[1]: https://docs.rs/tracing-subscriber/0.3.17/tracing_subscriber/filter/struct.EnvFilter.html#directives
[2]: https://docs.rs/env_logger/0.10.0/env_logger/#enabling-logging
```


rust/noosphere-core/src/tracing.rs
```diff
@@ -1,27 +1,390 @@
...
+    "noosphere_storage",
+    "noosphere_sphere",
+    "noosphere_into",
+    "noosphere_gateway",
+    "noosphere_collections",
+    "noosphere_cli",
+    "noosphere_car",
+    "noosphere_api",
+    "noosphere_ns",
+    "tower_http",
```
jsantell, line 20: Any ideas on keeping this in sync with new crates? Could be easy to miss

rust/noosphere-core/src/tracing.rs
```diff
@@ -1,27 +1,390 @@
...
+    /// [2]: https://docs.rs/env_logger/0.10.0/env_logger/#enabling-logging
+    pub fn initialize_tracing(noosphere_log: Option<NoosphereLog>) {
+        INITIALIZE_TRACING.call_once(|| {
+            if let Err(error) = initialize_tracing_subscriber(noosphere_log) {
+                println!("Failed to initialize tracing: {}", error);
+            }
+        });
+    }
+
+    fn initialize_tracing_subscriber(noosphere_log: Option<NoosphereLog>) -> anyhow::Result<()> {
```
jsantell, line 242: consider: returning `()`, making this function's signature match its infallibility

rust/noosphere-core/src/tracing.rs
```diff
@@ -1,27 +1,390 @@
...
+                .init(),
+        };
+
+        Ok(())
+    }
+
+    /// A formatter designed to make `INFO` events display as closely as
+    /// possible to regular `println` output, while allowing arbitrary other
+    /// formatting for all other event types
+    struct NoosphereMinimalFormatter<F, S, N>(F, PhantomData<S>, PhantomData<N>)
```
jsantell, line 331: Very nice! Big blocker to switching over to smarter logs has been the verbosity of the default formatter

rust/noosphere-into/src/into/html/sphere.rs
```diff
@@ -215,9 +215,9 @@ pub mod tests {
 
         let html = std::str::from_utf8(&bytes).unwrap();
 
-        println!();
-        println!("{}", html);
-        println!();
+        info!("");
+        info!("{}", html);
+        info!("");
```
jsantell, line 220: ~Could get weird with other formatters? Could `"\n{}\n"` work?~ nevermind this is a test

rust/noosphere-sphere/src/context.rs
```diff
@@ -543,10 +543,46 @@ pub mod tests {
         assert_eq!(name.as_str(), "c");
     }
 
+    #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
+    #[cfg_attr(not(target_arch = "wasm32"), tokio::test)]
+    async fn it_can_traverse_a_sequence_of_petnames_one_at_a_time() {
```
jsantell, line 548: Related to #320?

rust/noosphere/src/ffi/tracing.rs
```diff
@@ -0,0 +1,65 @@
...
+    /// Equivalent to minimal format / `DEBUG` filter
+    Chatty = NOOSPHERE_LOG_CHATTY,
+    /// Equivalent to verbose format / `DEBUG` filter
+    Informed = NOOSPHERE_LOG_INFORMED,
+    /// Equivalent to pretty format / `DEBUG` filter
+    Academic = NOOSPHERE_LOG_ACADEMIC,
+    /// Equivalent to verbose format / `TRACE` filter
+    Tiresome = NOOSPHERE_LOG_TIRESOME,
+    /// Equivalent to pretty format / `TRACE` filter
+    Deafening = NOOSPHERE_LOG_DEAFENING,
```
jsantell, line 30: While examples should this working as expected, enums exported via safer_ffi for error constants did not put a `_` between enum name and value -- does header generate `NS_NOOSPHERE_LOG_SILENT` or `NS_NOOSPHERE_LOGSILENT`?

also, consider: `ns_noosphere`/`NsNoosphere` refers to the top-level context object, which logs can be thought of as related. Similarly, maybe `ns_log` (e.g. `NS_LOG_SILENT`) is sufficient (though may be fighting some Apple **N**ame **S**paces), though not unique to this API

rust/noosphere/src/ffi/tracing.rs
```diff
@@ -0,0 +1,65 @@
...
+const NOOSPHERE_LOG_SILENT: u32 = 0;
+const NOOSPHERE_LOG_BASIC: u32 = 1;
+const NOOSPHERE_LOG_CHATTY: u32 = 2;
+const NOOSPHERE_LOG_INFORMED: u32 = 3;
+const NOOSPHERE_LOG_ACADEMIC: u32 = 4;
+const NOOSPHERE_LOG_TIRESOME: u32 = 5;
+const NOOSPHERE_LOG_DEAFENING: u32 = 6;
+
+#[derive_ReprC(rename = "ns_noosphere_log")]
+#[repr(u32)]
```
jsantell, line 13: Could be `u8`/`uint8_t`

swift/Tests/SwiftNoosphereTests/NoosphereTests.swift
```diff
@@ -8,6 +8,8 @@ final class NoosphereTests: XCTestCase {
         // reading from swift works as intended
         let noosphere = ns_initialize("/tmp/foo", "/tmp/bar", nil, nil)
 
+        ns_tracing_initialize(NS_NOOSPHERE_LOG_CHATTY.rawValue)
```
jsantell, line 11: :eyes: looks like it does format the enum values correctly! I'll need to look into https://github.com/subconsciousnetwork/noosphere/pull/340/files#diff-3359c09e1f3a2f311ceffbe341da1c49bf92026785af11d7bd1556aa28722d48R10 to see what's different here (u8 v u32?)

rust/noosphere/src/ffi/tracing.rs
```diff
@@ -0,0 +1,65 @@
...
+#[ffi_export]
+/// Configuration presets for Noosphere log behavior
```
jsantell, line 15: As this gets rendered via doxygen, and we use ints rather than enum for interface, consider mentioning this is used in `ns_tracing_initialize()` (can write as `"..log behavior, used in ns_tracing_initialize."`, no formatting or parens needed)

swift/Tests/SwiftNoosphereTests/NoosphereTests.swift
```diff
@@ -8,6 +8,8 @@ final class NoosphereTests: XCTestCase {
...
+
```
jsantell, line 12: Could drop this line in `c/example/main.c` as well!

jsantell (approved):
LGTM! Thrilled to reconcile our dev logs and CLI feedback, and fix the rough edges of wrangling logs in a web of crates. Some comments and considerations

swift/Tests/SwiftNoosphereTests/NoosphereTests.swift
jsantell: Maybe an issue with a version of safer_ffi on a different machine? No idea why my other machine was rendering like that. Fixed in #340. Disregard!!

rust/noosphere-core/src/tracing.rs
cdata: No, open to suggestions

rust/noosphere-core/src/tracing.rs
cdata: It's not infallible, and this let's us roll all the errors up into one at the call site.

rust/noosphere-sphere/src/context.rs
cdata: Yes, I wrote this change while investigating that and since I had the test I included here.

rust/noosphere/src/ffi/tracing.rs
cdata: In Swift, [`CInt` is `i32`](https://developer.apple.com/documentation/swift/cint) so this makes saves from some conversions. I would use `u8` if it was similarly ergonomic to do so.

cdata: My face as I realize the Rust side is all unsigned :facepalm:

rust/noosphere/src/ffi/tracing.rs
cdata: The related enum is literally `NoosphereLog`, and is a counterpart to the accepted values of the `NOOSPHERE_LOG` environment variable. `NsNoosphere` is the top-level context object, indeed.

I would maybe consider adding a prefix to the constants `NS_NOOSPHERE_LOG_SILENT`. But, to my eyes truncating e.g., `NOOSPHERE_LOG_*` to `NS_LOG_*` is moving away from legibility because it means we specify the log level with different language in the C FFI compared to both the Rust API and the environment variable.

rust/noosphere/src/ffi/tracing.rs
cdata: After more exploration, filed #345

PR #343, opened 2023-05-01
chore: Fix deprecation notes for updated libp2p_noise crate, update dependencies

jsantell:
One of the sub-libp2p packages (libp2p_noise) has updated. Not sure why I'm getting these notices without explicitly having updated the subpackage (which was done in order to reconcile the warnings in the first place, in order to have the suggestions available), similar to the previous deprecation warning fix by resetting the dependencies.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -164,9 +164,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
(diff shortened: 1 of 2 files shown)


cdata (approved):
LGTM

PR #354, opened 2023-05-03
feat: consider creating a new key with an empty string an error. Fixes #331

### LINKED ISSUE #331: Noosphere should reject key names that are empty strings
Currently, Noosphere incorrectly allows a key name to be a zero-length string when initializing a sphere. This leads to confusing errors when trying to look up the key name later. It should be explicitly disallowed.

jsantell:


cdata (approved):
Nice, thanks for this!

PR #355, opened 2023-05-03
feat: add c/c++ vscode settings and clang-format manifest. Fixes #344.

### LINKED ISSUE #344: Commit C/C++ configurations to source
For vscode users at least, it'd be nice to commit explicit values for include paths for the examples, and to reduce PR noise, a consistent formatter/configuration.

jsantell:
Other possibilities:

* Setting C/C++ formatter in VSCode settings. By default, `file` (e.g. `.clang-format`) is used by VSCode, but we can specify a fallback if this isn't found (`C_Cpp.clang_format_fallbackStyle`). This wouldn't handle non-vscode editors.
* Alternative clang format defaults are available, e.g. Mozilla, Chromium, LLVM, etc. and can be generated via `clang-format -style=Google -dump-config > .clang-format`. I don't have much preference here.

Without this or similar, changing C code reformats based off of local configurations.

### DIFF
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -1,0 +1,16 @@
+{
+    "configurations": [
+        {
+            "name": "Linux",
+            "includePath": [
+                "${workspaceFolder}/**"
+            ],
+            "defines": [],
+            "compilerPath": "/usr/bin/clang",
+            "cStandard": "c17",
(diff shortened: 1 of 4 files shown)


.clang-format
```diff
@@ -0,0 +1,226 @@
+---
```
subconscious-ops, line 1: Can we move this down into the `c` folder?

PR #357, opened 2023-05-03
Update orb-ns for cluster

jsantell:
* remove mutex from NameSystem's APIServer
* Add logs for RUST_LOG, NOOSPHERE_LOG in image startup scripts
* add `orb` and `orb_ns` in default tracing

Added here, but not immediate relevant:
* support defining partial DhtConfig values via TOML: ultimately we want to expose these knobs, but turns out libp2p's `query_timeout` is not the overall request timeout that we want currently (as a "request" is comprised of several "queries")
* improve `NsRecord`'s Debug formatting: useful, but really need to fix the debug representation of libp2p's `Record`
* remove `tracing_subscriber` from noosphere_ns, causing wild logging results (unfortunately did not fix tracing weirdness in noosphere_ns)

### DIFF
--- a/images/orb/start.sh
+++ b/images/orb/start.sh
@@ -23,4 +23,7 @@ if ! [ -z "$NS_API" ]; then
 	ARGS="${ARGS} --name-resolver-api ${NS_API}"
 fi
 
+echo "RUST_LOG=${RUST_LOG}"
+echo "NOOSPHERE_LOG=${NOOSPHERE_LOG}"
+
 orb serve $ARGS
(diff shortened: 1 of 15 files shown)


images/orb/start.sh
```diff
@@ -23,4 +23,7 @@ if ! [ -z "$NS_API" ]; then
 	ARGS="${ARGS} --name-resolver-api ${NS_API}"
 fi
 
+echo "RUST_LOG=${RUST_LOG}"
+echo "NOOSPHERE_LOG=${NOOSPHERE_LOG}"
```
cdata, line 27: It's too bad this isn't easy to find in the Nomad UI...

cdata (approved):
LGTM

PR #360, opened 2023-05-04
feat: Enable expired yet valid records in the name system. Update to ucan 0.2.0

jsantell:
Thanks to @cdata and https://github.com/ucan-wg/rs-ucan/pull/83, we can construct a ProofChain with a valid, yet expired, record. Add a test, and change the behavior of other tests: this is now allowed!


cdata (approved):
Nice!

PR #373, opened 2023-05-09
fix: Removed petnames stay removed

### LINKED ISSUE #350: Deleted petnames re-added after gateway sync
Related Subconscious bug:

https://github.com/subconsciousnetwork/subconscious/issues/569

cdata:
This fixes a bug where a petname that was removed locally would seem to be re-added to the address book after a gateway. Most of the change is in the form of an integration test that tests for the expected the behavior.

Fixes #350 
Fixes https://github.com/subconsciousnetwork/subconscious/issues/569


jsantell (approved):
LGTM!

PR #376, opened 2023-05-09
chore: Commit and tag Package.swift during release

### LINKED ISSUE #135: Evaluate automatic updates to `Package.swift`
As of #131 , we have a Swift package and a bunch of Github Actions machinery for producing the artifacts it needs. However, at this time the `Package.swift` has to be updated by hand after new releases of the `noosphere` crate are made.

We should evaluate doing automatic updates of the `Package.swift` after `noosphere` crate releases, as well as the possibility of tagging releases for the Swift Package generally.

cdata:
This change adds a step to our release that commits and tags `Package.swift` so that it points to the latest Noosphere release. The latest Noosphere release is tagged as `noosphere-$VERSION`, and the corresponding SwiftNoosphere release is tagged as `swift-noosphere-$VERSION`.

Fixes #135


jsantell (approved):
:tada:

PR #382, opened 2023-05-11
feat: validate petnames and slugs disallowing an empty string

jsantell:


rust/noosphere-sphere/src/content/write.rs
```diff
@@ -12,6 +12,14 @@ use async_trait::async_trait;
 
 use crate::{AsyncFileBody, SphereContentRead};
 
+fn validate_slug(slug: &str) -> Result<()> {
+    if slug.is_empty() {
+        Err(anyhow!("slug must not be empty."))
```
cdata, line 17: Capitalize `Slug`

rust/noosphere-sphere/src/content/write.rs
```diff
@@ -91,6 +99,7 @@ where
         additional_headers: Option<Vec<(String, String)>>,
     ) -> Result<Cid> {
         self.assert_write_access().await?;
+        validate_slug(slug)?;
```
cdata, line 102: Should also guard `link_raw` above.

cdata (commented):
One thing that strikes me, reading through this: there are a number of places where "slug" or "petname" is the intended argument, and these are really stringly typed values. For example, if we validate the slug format on writes, we should probably also do it on reads.

We've previously adapted to stringly types by making concrete wrapper types for them. Maybe this is a case where we need a `Slug` and `Petname` wrapper type that can only be initialized on a code path that performs validation. Then, we can change all the related signatures to use this wrapper type instead.

WDYT?

jsantell:
> One thing that strikes me, reading through this: there are a number of places where "slug" or "petname" is the intended argument, and these are really stringly typed values. For example, if we validate the slug format on writes, we should probably also do it on reads.
> 
> We've previously adapted to stringly types by making concrete wrapper types for them. Maybe this is a case where we need a `Slug` and `Petname` wrapper type that can only be initialized on a code path that performs validation. Then, we can change all the related signatures to use this wrapper type instead.
> 
> WDYT?

Rationale behind guarding writes and not reads: we could introduce an invalid petname/slug somewhere, and at least we'll be able to query its existence and read/delete it (even in tests). On the contrast, assuming that all slug/petnames are valid internally in non-writes could cause an issue.

Because the values aren't (yet?) restrictive (only guarding against empty strings) or have multiple interpretations (ucan/nsrecord/jwt), I'm bearish on new String-types for Petname/Slug

cdata:
Filed https://github.com/subconsciousnetwork/noosphere/issues/385 for potential future exploration here.

Issue #383, opened 2023-05-12
Petnames assigned to identities: use history to compute delta changes


cdata:
We currently compute the petnames assigned to a DID on-demand in the most naive way possible: by crawling the address book and computing the inverse mapping of all the petname keys. We then cache the results for that version of the sphere. It would be better if we could checkpoint and update the cache incrementally based on the revisions of the sphere since the last checkpoint.

PR #384, opened 2023-05-12
feat: Get petnames assigned to a DID for a sphere

### LINKED ISSUE #358: C FFI to get all petnames for a DID from a sphere
We need the ability to get all petnames for a DID, given a sphere. The petnames should be limited to those that are assigned to the given DID within the address book of the given sphere only.

This feature poses the following challenges:

 - Keeping the books within the sphere data structure would significantly complicate the data structure
 - Not keeping the books would result in costly traversals every time we move to derive the answer

With that in mind, we should probably satisfice by taking the slow approach by default and caching the results in `SphereDb`. A purpose-built storage mechanism should be added to `SphereDb` that allows us to store multi-dimensional keys. Today we need two dimensions: DID(origin sphere) x DID(peer sphere). It's probably ideal to conceive of this kind of storage as N dimensional, but the base case is two.

cdata:
This introduces a built-in facility for getting all of the petnames assigned to a DID at a given version of a sphere. The discovery is limited to petnames and DIDs that are in the current sphere's address book.

Since petnames are encoded `petname -> DID` in the data structure, we compute the inverse assignments on-demand and cache the results for a given version of the sphere. This represents a basic optimization over an expensive operation. In time, we may be able to optimize by only computing updates based on sphere history deltas (https://github.com/subconsciousnetwork/noosphere/issues/383).

Also included in this change is the corresponding callback-based C FFI, and a Swift helper and example test.

Fixes #358

### DIFF
--- a/rust/noosphere-sphere/src/cursor.rs
+++ b/rust/noosphere-sphere/src/cursor.rs
@@ -161,32 +161,11 @@ where
     }
 }
 
-#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
-#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
-impl<'a, C, K, S> HasSphereContext<K, S> for &'a SphereCursor<C, K, S>
-where
-    C: HasSphereContext<K, S>,
-    K: KeyMaterial + Clone + 'static,
-    S: Storage + 'static,
-{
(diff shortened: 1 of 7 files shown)


jsantell (dismissed):
LGTM!

PR #387, opened 2023-05-12
chore: disallow a sphere adding itself via petname. Fixes #293

### LINKED ISSUE #293: Following yourself creates an infinite loop
> cdata: This is bringing to mind that we don't stop you from following yourself, and that this could cause an infinite loop as you update your address book to point to your latest sphere version, which causes a yet newer version to be published to the name system, etc.

> cdata: Yah, I think Noosphere needs to special case this so that your own sphere's DID always resolves to its own local tip or something

jsantell:


rust/noosphere-sphere/src/petname/write.rs
```diff
@@ -6,9 +6,15 @@ use ucan::{crypto::KeyMaterial, store::UcanJwtStore, Ucan};
 
 use crate::{internal::SphereContextInternal, HasMutableSphereContext, SpherePetnameRead};
 
-fn validate_petname(petname: &str) -> Result<()> {
+fn validate_petname(
```
cdata, line 9: Can we simplify this to disallowing any `did:` petname? That is probably always a bad idea because the user would be creating a misleading alias. And, it would save us from reaching into the `SphereContext` just to get a value to validate against.

cdata (approved):
Nice!

PR #396, opened 2023-05-17
feat: return `None` instead of throwing in `ns_sphere_petname_resolve()` when a petname cannot be resolved.

### LINKED ISSUE #393: `ns_sphere_petname_resolve` gives error when no link record has been resolved
Currently the C FFI `ns_sphere_petname_resolve` will give the caller an error if they try to resolve a petname to a sphere version when no link record has been found for the DID that the petname is assigned to. However, the intended behavior of the function is to return a null pointer with _no_ error in this case. We should fix the FFI call to reflect these semantics.

jsantell:
Fixes #393


cdata (approved):
Great job on this return thing

PR #399, opened 2023-05-18
feat: Promote NsRecord into a core object

### LINKED ISSUE #395: Migrate NsRecord to a core type
`NsRecord`, a wrapper around Ucans for name system publishing, is used by the name system, as well as stored in spheres as JWTs and used as a core data type throughout noosphere. We should generalize NsRecord and bring mainly its validation to `noosphere-core`, probably related to LinkRecord, to avoid cyclical dependencies and bringing in a whole name system for crates that want to validate a record.

jsantell:
... by extending noosphere_core::data::LinkRecord with its functionality, and removing the noosphere_ns dependency on validating records. Fixes #395


rust/noosphere-core/src/data/address.rs
```diff
@@ -65,13 +74,88 @@ pub struct LinkRecord(Jwt);
...
     /// Parse the wrapped [Jwt] as a [Ucan] and looks for the referenced pointer
     /// to some data in IPFS (via a [Cid] in the `fct` field).
-    pub async fn dereference(&self) -> Option<Cid> {
-        let token = &self.0;
-        let ucan = match Ucan::try_from(token.to_string()) {
-            Ok(ucan) => ucan,
-            _ => return None,
+    pub fn dereference(&self) -> Option<Cid> {
+        match Ucan::try_from(self) {
+            Ok(ucan) => LinkRecord::link(&ucan),
```
cdata, line 79: A verb might be nice here e.g., `LinkRecord::find_link` or `extract_link` or something..

rust/noosphere-core/src/data/address.rs
```diff
@@ -65,13 +74,88 @@ pub struct LinkRecord(Jwt);
...
+        }
+    }
+
+    /// Validates the [Ucan] token as a [LinkRecord], ensuring that
+    /// the sphere's owner authorized the publishing of a new
+    /// content address. Returns an `Err` if validation fails.
+    pub async fn validate<S: UcanJwtStore>(
+        token: &Ucan,
+        store: &S,
+        opt_did_parser: Option<&mut DidParser>,
```
cdata, line 90: I've basically given up on including this in signatures. We should just instantiate `DidParser` at the point of use; it's just going to be removed when we update `rs-ucan` for the v0.10.0 UCAN spec.

rust/noosphere-core/src/data/address.rs
```diff
@@ -65,13 +74,88 @@ pub struct LinkRecord(Jwt);
...
+    pub fn dereference(&self) -> Option<Cid> {
+        match Ucan::try_from(self) {
+            Ok(ucan) => LinkRecord::link(&ucan),
+            _ => None,
+        }
+    }
+
+    /// Validates the [Ucan] token as a [LinkRecord], ensuring that
+    /// the sphere's owner authorized the publishing of a new
+    /// content address. Returns an `Err` if validation fails.
```
cdata, line 86: It would probably be good to note the context-specific rules for expiry validation here in this comment.

rust/noosphere-core/src/data/address.rs
```diff
@@ -99,11 +183,37 @@ impl LinkRecord {
...
+    ///     with: With::Resource {
+    ///         kind: Resource::Scoped(SphereReference {
+    ///            did: identity.to_owned(),
+    ///         }),
+    ///     },
+    ///     can: SphereAction::Publish,
+    /// };
+    /// assert_eq!(LinkRecord::generate_capability(&identity), expected_capability);
+    /// ```
+    pub fn generate_capability(identity: &str) -> Capability<SphereReference, SphereAction> {
```
cdata, line 207: A generic wrapper for capability generation might be more appropriate e.g., `generate_capability(sphere: &Did, action: SphereAction)`, and we can stick it in `noosphere-core::authority`.

rust/noosphere-core/src/data/address.rs
```diff
@@ -65,13 +74,88 @@ pub struct LinkRecord(Jwt);
...
+    }
+
+    /// Returns true if the [Ucan] token is currently publishable
+    /// within the bounds of its expiry/not before time.
+    pub fn has_publishable_timeframe(token: &Ucan) -> bool {
+        !token.is_expired(None) && !token.is_too_early()
+    }
+
+    /// The DID key of the sphere that this record maps.
+    pub fn identity(token: &Ucan) -> &str {
```
cdata, line 152: `sphere_identity`

cdata (changes requested):
This mostly looks good.

One thing that stood out to me is that _all_ of `LinkRecord`'s implementation is static in order to work around the fact that what it really wants to be operating on is a `Ucan` (but what it has available is an encoded `Jwt`).

It seems like there are two possible revisions to this change that are worth choosing from:

 1. Refactor `LinkRecord` to hold a `Ucan` internally (_not_ a `Jwt`), or....
 2. Add yet-another-wrapper that takes a `LinkRecord` and internally also holds a corresponding `Ucan`
 
After applying one of these approaches, the static methods can be re-imagined as instance methods without extra cost required to parse a `Ucan` for every call.

jsantell:
Reconfigured to wrap a Ucan, not dissimilar from NsRecord; updating accordingly

* Promote `noosphere_core::authority::generate_capability(identity: &str, action: SphereAction)`
* Remove optional `DidParser` from `LinkRecord::validate` method
* Updated some traits accordingly, similar to what NsRecord needed as a Ucan wrapper (rather than a Jwt wrapper)
* Updated method names (with no static methods, `dereference` was redundant with `link`, renamed to `get_link()`)

rust/noosphere-core/src/data/address.rs
```diff
@@ -99,15 +153,60 @@ impl LinkRecord {
...
         }
-
         warn!("No facts contained a link!");
-
         None
     }
+
+    /// Return a base64-encoded string serialization of the
+    /// underlying [Ucan].
+    pub fn encode(&self) -> Result<String> {
```
cdata, line 162: I would say just make this `Deref<Target = Ucan>` (which it already is, I think?) and invoke `encode` directly.

Alternatively, it this is important than we should eat our veggies and return a `Result<Jwt>` here.

rust/noosphere-core/src/data/address.rs
```diff
@@ -99,15 +153,60 @@ impl LinkRecord {
...
+        self.0.encode()
+    }
+
+    /// Return a base64-encoded [Vec<u8>] of the underlying [Ucan].
+    pub fn as_bytes(&self) -> Result<Vec<u8>> {
```
cdata, line 167: This seems a little redundant; you can just call `as_bytes`  on any string in Rust

rust/noosphere-core/src/data/address.rs
```diff
@@ -99,15 +153,60 @@ impl LinkRecord {
...
+        let s = String::deserialize(deserializer)?;
+        let record = LinkRecord::try_from(s).map_err(de::Error::custom)?;
+        Ok(record)
+    }
+}
+
+/// [LinkRecord]s compare their [Jwt] representations
+/// for equality. If a record cannot be encoded as such,
+/// they will not be considered equal to any other record.
+impl PartialEq for LinkRecord {
```
cdata, line 196: This is a pretty expensive equality check. Would you be willing to file an issue for us to implement `PartialEq` for [`Ucan` et al](
https://github.com/ucan-wg/rs-ucan/blob/main/ucan/src/ucan.rs#L20-L42C18) so that we can avoid the encoding cost?

cdata (commented):
Just a little more cleanup requested :pray:

rust/noosphere-core/src/data/address.rs
jsantell: Agreed, will do!

cdata (approved):
Nice work!

PR #409, opened 2023-05-30
fix!: Enable incremental sphere replication

cdata:
- Most use of `Cid` where it refers to a sphere has been refactored as `Link<MemoIpld>`
 - Gateway Fetch API block `Bundle` is replaced with a CAR stream
 - Peer `LinkRecord` proofs are replicated along with the rest of a given sphere's blocks
 - `MemoIpld` now records a Lamport timestamp to establish a causal order between related versions of its body content
 - `MemoIpld` signatures previously referred to a proof by CID, but now inline an invoked UCAN JWT
 - Traversal logic is moved out of `SphereContext`, and the core of it is defined as part of `Sphere`; additional business logic pertaining to replication is implemented as a trait over `HasSphereContext`


rust/noosphere-core/src/data/memo.rs
```diff
@@ -311,4 +348,34 @@ mod test {
...
+
+        let first_memo = memos.get(0).unwrap();
+        let third_memo = memos.get(2).unwrap();
+        let fourth_memo = memos.get(3).unwrap();
+        let fifth_memo = memos.get(4).unwrap();
+
+        assert!(third_memo.lamport_order() < fourth_memo.lamport_order());
+        assert!(fifth_memo.lamport_order() > fourth_memo.lamport_order());
+        assert!(first_memo.lamport_order() < third_memo.lamport_order());
+        assert!(first_memo.lamport_order() < third_memo.lamport_order());
```
jsantell, line 377: Dupe line?

rust/noosphere-core/src/view/sphere.rs
```diff
@@ -606,8 +786,68 @@ impl<S: BlockStore> Sphere<S> {
...
+
+            // Discover the intended audience of the UCAN
+            let credential = did_parser.parse(ucan.audience())?;
+
+            // Verify the audience signature of the body CID
+            credential.verify(&memo.body.to_bytes(), &signature).await?;
+
+            // Check the proof's provenance and that it enables the signer to sign
+            let ucan_store = UcanStore(self.store.clone());
+            let proof = ProofChain::from_ucan(ucan, None, &mut did_parser, &ucan_store).await?;
```
jsantell, line 827: Do we need to consider publication time validity, like link records?

rust/noosphere-gateway/src/route/push.rs
```diff
@@ -104,6 +103,11 @@ where
     pub async fn invoke(mut self) -> Result<PushResponse, PushError> {
         debug!("Invoking gateway push...");
 
+        // let sphere_context = self.sphere_context.sphere_context_mut().await?;
+
+        // self.verify_history(&mut sphere_context).await?;
+        // self.incorporate_history(&mut sphere_context).await?;
+        // self.synchronize_names(&mut sphere_context).await?;
```
jsantell, line 110: nit: commented lines

rust/noosphere-gateway/src/route/replicate.rs
```diff
@@ -62,5 +73,302 @@ where
...
+                return Ok(StreamBody::new(Box::pin(car_stream(
+                    vec![memo_version],
+                    memo_history_stream(store, &memo_version.into(), Some(&since)),
+                ))));
+            } else {
+                error!("Suggested version {since} is not a valid ancestor of {memo_version}");
+                return Err(StatusCode::BAD_REQUEST);
+            }
+        } else {
+            warn!(
```
jsantell, line 103: If `since_memo == latest_memo`, there should be nothing to sync I believe, which triggers this slightly inaccurate warning

rust/noosphere-gateway/src/route/replicate.rs
```diff
@@ -62,5 +73,302 @@ where
...
+                "Client's version {} is newer than {}; can't stream incrementally",
```
jsantell, line 104: If `since_memo >= latest_memo`, we fallback to the full replication -- I'd expect an empty success response in this case e.g. "no work to do", as opposed to full replication as if `since` wasn't provided

rust/noosphere-sphere/Cargo.toml
```diff
@@ -40,6 +40,9 @@ libipld-cbor = { workspace = true }
 bytes = "^1"
 serde_json = { workspace = true }
 serde = { workspace = true }
+thiserror = { workspace = true }
+# strum = { workspace = true }
```
jsantell, line 44: nit: comments

rust/noosphere-sphere/src/context.rs
```diff
@@ -254,20 +111,30 @@ where
...
-    }
-
     /// The identity of the sphere
     pub fn identity(&self) -> &Did {
         &self.sphere_identity
     }
 
+    /// The identity of the counterpart sphere in use during this session, if
+    /// any; note that this will cause a request to be made to a gateway if no
+    /// handshake has yet occurred.
```
jsantell, line 121: Good call out on methods that use the network

rust/noosphere-sphere/src/context.rs
```diff
@@ -254,20 +111,30 @@ where
...
+    pub async fn counterpart_identity(&self) -> Result<Did> {
+        Ok(self.client().await?.session.sphere_identity.clone())
+    }
+
+    /// The identity of the counterpart sphere in use during this session, if
```
jsantell, line 126: nit: s/counterpart/gateway

rust/noosphere-sphere/src/has.rs
```diff
@@ -105,7 +97,10 @@ where
     /// new version [Cid] of the sphere is returned. This method must be invoked
```
jsantell, line 97: followup: I bet there's a handful of `[Cid]` usage in docs now that should be `Link<MemoIpld>`

rust/noosphere-sphere/src/sync/recovery.rs
```diff
@@ -0,0 +1,5 @@
+#[derive(Debug)]
+pub enum SyncRecovery {
```
jsantell, line 2: Could use some light docs, unless explained by other functions

rust/noosphere-sphere/src/context.rs
```diff
@@ -254,20 +111,30 @@ where
...
-
     /// The identity of the sphere
     pub fn identity(&self) -> &Did {
         &self.sphere_identity
     }
 
+    /// The identity of the counterpart sphere in use during this session, if
+    /// any; note that this will cause a request to be made to a gateway if no
+    /// handshake has yet occurred.
+    pub async fn counterpart_identity(&self) -> Result<Did> {
```
jsantell, line 122: How is `counterpart_identity()` used? Is this the gateway's counterpart, e.g. *this* sphere's identity?

jsantell (dismissed):
Great work tackling this challenge!

rust/noosphere-sphere/src/has.rs
cdata: Tried to fix as many of these as I could find :+1:

rust/noosphere-gateway/src/route/replicate.rs
cdata: Tweaked `is newer` to `is not older`; burden is on the caller to provide expected inputs in this case, so it's correct to warn.

rust/noosphere-gateway/src/route/replicate.rs
cdata: This is err'ing on the side of caution. Caller wants a version of the sphere, but does not realize that the requested version was older than than the one they already have (and in many cases they can't know it because they don't even have the memo they are requesting).

rust/noosphere-cli/tests/peer_to_peer.rs
```diff
@@ -244,3 +236,274 @@ async fn synchronize_petnames_as_they_are_added_and_removed() -> Result<()> {
...
+
+            assert_eq!(cid, sphere_3_newest_version);
+
+            let cursor = SphereCursor::latest(Arc::new(ctx.sphere_context().await?.clone()));
+            let sphere_1_sphere_3_cursor = cursor
+                .traverse_by_petnames(&["pair_3".to_string()])
+                .await?
+                .unwrap();
+
+            let mut file = sphere_1_sphere_3_cursor.read("baz").await?.unwrap();
```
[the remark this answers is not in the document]
cdata: The intention is to ensure both sides are okay. We got the latest changes, and we also got the old version as expected.

rust/noosphere-sphere/src/context.rs
cdata: Strictly speaking the counterpart is relative to your sphere. For the gateway, the counterpart is the user's sphere. For the user, the counterpart is the gateway's sphere.

That said, this method isn't in use and its implementation only works for the user, so we should just remove it.

rust/noosphere-sphere/src/context.rs
cdata: This is a bad mental model for implementation that lives on a sphere context. A gateway will make use of a sphere context, but it does not itself have a gateway to talk to.

This is confounded by the fact that sphere context's have a built-in client for the gateway API, so let's leave as is for now. But, we need to change this soon because it's a footgun when used within the gateway (or in the future, within a Geist).

rust/noosphere-core/src/view/sphere.rs
cdata: Good catch, yes.

cdata: (BTW, this is not new implementation; this was moved into `Sphere` but used to be a stand-alone function used in tests)

PR #412, opened 2023-06-08
fix: Resolve petnames in correct order

### LINKED ISSUE #411: Changed petname is assigned to wrong (previous) DID
See: https://github.com/subconsciousnetwork/subconscious/issues/608

When we assign a petname, then unassign it, then re-assign it to a different DID, all while sync-ing with a gateway, we end up in a situation where the petname is assigned to the first DID it was assigned to, which is incorrect. We expect it to be assigned to the most recent DID it was assigned to.

cdata:
After investigating https://github.com/subconsciousnetwork/subconscious/issues/608, a few issues were uncovered related to how petnames are resolved in the gateway. Most importantly, the gateway was suffering from the effects of a race condition when adopting petnames. But, addressing this uncovered a more subtle problem: it is not safe for us to blindly adopt link records that don't match up with the petname we are adopting them against. For example, consider the following sequence of events:

1. A petname is assigned to a DID
2. During sync, the gateway kicks off a parallel job to resolve the petname
3. Meanwhile, we unassign the petname
4. We sync, the gateway takes no action (no new names to resolve)
5. Then, the original resolve job finishes and comes back with a record

Prior to this change, we would just adopt the record, and eventually so would the client. After all, record adoption is not able to disambiguate between between a new record being added and a race condition like the one described above.

Now, it is mandatory that a petname be assigned to a DID before a record be adopted. And, `adopt_petname` has been renamed `set_petname_record` to better communicate the semantics. The old method remains but is marked as deprecated.

Fixes #411

### DIFF
--- a/rust/noosphere-cli/tests/peer_to_peer.rs
+++ b/rust/noosphere-cli/tests/peer_to_peer.rs
@@ -194,14 +194,27 @@ async fn synchronize_petnames_as_they_are_added_and_removed() -> Result<()> {
 
     let mut base_pair = SpherePair::new("BASE", &ipfs_url, &ns_url).await?;
     let mut other_pair = SpherePair::new("OTHER", &ipfs_url, &ns_url).await?;
+    let mut third_pair = SpherePair::new("THIRD", &ipfs_url, &ns_url).await?;
 
     base_pair.start_gateway().await?;
     other_pair.start_gateway().await?;
+    third_pair.start_gateway().await?;
 
     let other_pair_id = other_pair.client.identity.clone();
     let other_version = other_pair
         .spawn(|mut ctx| async move {
             ctx.write("foo", "text/plain", "bar".as_ref(), None).await?;
+            let version = ctx.save(None).await?;
+            ctx.sync(SyncRecovery::Retry(3)).await?;
+            wait(1).await;
+            Ok(version)
+        })
+        .await?;
+
+    let third_pair_id = third_pair.client.identity.clone();
+    let third_version = third_pair
+        .spawn(|mut ctx| async move {
+            ctx.write("bar", "text/plain", "baz".as_ref(), None).await?;
             let version = ctx.save(None).await?;
             ctx.sync(SyncRecovery::Retry(3)).await?;
             wait(1).await;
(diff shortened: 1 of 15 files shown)


rust/noosphere-cli/tests/peer_to_peer.rs
```diff
@@ -227,8 +240,27 @@ async fn synchronize_petnames_as_they_are_added_and_removed() -> Result<()> {
             ctx.set_petname("thirdparty", None).await?;
             ctx.save(None).await?;
             ctx.sync(SyncRecovery::Retry(3)).await?;
+            wait(1).await;
+            ctx.sync(SyncRecovery::Retry(3)).await?;
```
jsantell, line 244: These extra syncs are suspicious, what changes are expected to be pushed or pulled from this second sync? These are of course integration tests fighting with real timing/race conditions where waiting will be necessary, but wondering what e.g. fails without these

rust/noosphere-sphere/src/petname/write.rs
```diff
@@ -24,17 +24,23 @@ where
...
-    /// publishes a name record and grants sufficient authority from the
-    /// configured [Did] to the publisher.
+    /// Set the [LinkRecord] associated with a petname.  The [LinkRecord] must
+    /// resolve a valid UCAN that authorizes the corresponding sphere to be
+    /// published and grants sufficient authority from the configured [Did] to
+    /// the publisher. The audience of the UCAN must match the [Did] that was
+    /// most recently assigned the associated petname. Note that a petname
+    /// _must_ be assigned to the audience [Did] in order for the record to be
+    /// set.
+    async fn set_petname_record(&mut self, name: &str, record: &LinkRecord) -> Result<Option<Did>>;
```
jsantell, line 41: Might be worth mentioning WHY we can't just implement `set_petname_record()` as previously implemented, rather than a two-step process of `set_petname` + `set_petname_record` e.g. documenting the need for this change

jsantell (approved):
Clever solution to a gnarly problem, nice!

rust/noosphere-cli/tests/peer_to_peer.rs
cdata: The gateway does not resolve names synchronously. First sync: client pushes name entries to gateway. Second sync: client gets the gateway's resolved link records.

Two syncs simulates the intended flow.

rust/noosphere-sphere/src/petname/write.rs
cdata: You mean as a comment? Can you paint a picture of who that benefits?

jsantell: Me in 6 months thinking, "why doesn't `set_petname_record()` also just set a petname->identity mapping in one-go?", e.g. why it's mandatory that a petname must already be assigned before adoption, though thinking about it more, this is similar to rolling other verifications from the NS worker into the petname writers

rust/noosphere-cli/tests/peer_to_peer.rs
jsantell: Was missing some context here for syncing a *removed* entry, makes sense!

rust/noosphere-sphere/src/petname/write.rs
cdata: I'll add a code comment reminder :+1:

PR #415, opened 2023-06-13
feat: Authorize and revoke authorization APIs

### LINKED ISSUE #325: C FFI to authorize and revoke authorizations from a sphere to a DID
Currently we have the ability to authorize / revoke authorizations from a sphere to a DID, but we don't expose it in the C FFI. We should do this so that the Subconscious app and other FFI consumers can take advantage of this functionality.

cdata:
This change proposes high-level APIs to authorize keys and revoke authorizations for both `SphereContext` wrappers and the C FFI. In addition to the proposed APIs, a bunch of docs have been backfilled.

A provisional implementation of sphere "recovery" has also been introduced here. This recovery mechanism is intended for the case when the user loses an authorized credential (due to theft or otherwise) and wishes to complete reset all authorizations via their recovery mnemonic. However, given the provisional nature of this implementation, no C FFI is provided for this functionality at this time.

Fixes #325

### DIFF
--- a/rust/noosphere-core/src/authority/author.rs
+++ b/rust/noosphere-core/src/authority/author.rs
@@ -1,18 +1,19 @@
 use crate::{
     authority::{
-        generate_ed25519_key, Authorization, SphereAction, SphereReference, SPHERE_SEMANTICS,
-        SUPPORTED_KEYS,
+        generate_ed25519_key, Authorization, SphereAction, SPHERE_SEMANTICS, SUPPORTED_KEYS,
     },
     data::Did,
+    view::Sphere,
 };
 use anyhow::{anyhow, Result};
 use noosphere_storage::{SphereDb, Storage};
 use ucan::{
-    capability::{Capability, Resource, With},
     chain::ProofChain,
     crypto::{did::DidParser, KeyMaterial},
 };
 use ucan_key_support::ed25519::Ed25519KeyMaterial;
+
+use super::generate_capability;
 
 /// The level of access that a given user has to a related resource. Broadly,
 /// a user will always have either read/write access (to their own sphere) or
@@ -71,24 +72,40 @@ where
         sphere_identity: &Did,
         db: &SphereDb<S>,
     ) -> Result<Access> {
+        let author_did = Did(self.key.get_did().await?);
+
+        // Check if this author _is_ the root sphere authority (e.g., when performing surgery on
+        // the authority section of a sphere)
+        if &author_did == sphere_identity {
+            return Ok(Access::ReadWrite);
+        }
+
         if let Some(authorization) = &self.authorization {
-            let author_did = Did(self.key.get_did().await?);
-            let ucan = authorization.resolve_ucan(db).await?;
+            let ucan = authorization.as_ucan(db).await?;
 
             if ucan.audience() != author_did.as_str() {
                 return Ok(Access::ReadOnly);
             }
 
-            let read_write_capability = Capability {
-                with: With::Resource {
-                    kind: Resource::Scoped(SphereReference {
-                        did: sphere_identity.to_string(),
-                    }),
-                },
-                can: SphereAction::Push,
+            let sphere = Sphere::at(&db.require_version(sphere_identity).await?.into(), db);
+            match sphere.verify_authorization(authorization).await {
+                Ok(_) => (),
+                Err(error) => {
+                    warn!("Could not verify authorization: {}", error);
+                    return Ok(Access::ReadOnly);
+                }
             };
+
+            let read_write_capability = generate_capability(sphere_identity, SphereAction::Push);
+
             let mut did_parser = DidParser::new(SUPPORTED_KEYS);
-            let proof_chain = ProofChain::from_ucan(ucan, None, &mut did_parser, db).await?;
+            let proof_chain = match ProofChain::from_ucan(ucan, None, &mut did_parser, db).await {
+                Ok(proof_chain) => proof_chain,
+                Err(error) => {
+                    warn!("Could not construct a verified proof chain: {}", error);
+                    return Ok(Access::ReadOnly);
+                }
+            };
 
             let capability_infos = proof_chain.reduce_capabilities(&SPHERE_SEMANTICS);
 
@@ -107,6 +124,7 @@ where
 
 #[cfg(test)]
 mod tests {
+    use anyhow::Result;
     use noosphere_storage::{MemoryStorage, SphereDb};
     use ucan::crypto::KeyMaterial;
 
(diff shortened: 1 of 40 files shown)


rust/noosphere-sphere/src/authority/escalate.rs
```diff
@@ -0,0 +1,85 @@
...
+where
+    C: HasMutableSphereContext<K, S>,
+    K: KeyMaterial + Clone + 'static,
+    S: Storage + 'static,
+{
+    /// Given the recovery mnemonic for a sphere and a callback, invoke the
+    /// callback with a mutable [SphereContext] wrapper. The optional return
+    /// value of the callback should be the latest version of the sphere after
+    /// the work of the callback is performed
+    async fn with_root_authority<F, Fut>(&mut self, mnemonic: &Mnemonic, callback: F) -> Result<()>
```
jsantell, line 44: What's the rational for a callback here as opposed to returning a root sphere context? Isolating scope/access?

rust/noosphere-sphere/src/authority/write.rs
```diff
@@ -0,0 +1,410 @@
...
+            let ucan = authorization
+                .as_ucan(&UcanStore(sphere.store().clone()))
+                .await?;
+            *ucan.expires_at()
+        };
+
+        let mut signable = UcanBuilder::default()
+            .issued_by(&author.key)
+            .for_audience(identity)
+            .claiming_capability(&Capability {
```
jsantell, line 82: nit: `generate_capability()`

rust/noosphere/src/ffi/authority.rs
```diff
@@ -0,0 +1,125 @@
...
+use crate::{
+    error::NoosphereError,
+    ffi::{NsError, NsNoosphere, NsSphere},
+};
+
+#[ffi_export]
+/// @memberof ns_sphere_t
+///
+/// Authorize another key to manipulate this sphere, given a display name for
+/// the authorization and the DID of the key to be authorized.
```
jsantell, line 20: consider adding `public` e.g. `"given a public display name for the authorization"`

rust/noosphere-sphere/src/authority/write.rs
```diff
@@ -0,0 +1,410 @@
...
+where
+    C: HasSphereContext<K, S> + HasMutableSphereContext<K, S>,
+    K: KeyMaterial + Clone + 'static,
+    S: Storage + 'static,
+{
+    // TODO: We allow optional human-readable names for authorizations, but this
+    // will bear the consequence of leaking personal information about the user
+    // (e.g., a list of their authorized devices). We should encrypt these
+    // names so that they are only readable by the user themselves.
+    async fn authorize(&mut self, name: &str, identity: &Did) -> Result<Authorization> {
```
jsantell, line 63: Took me a bit to get this: `authorize()` authorizes an identity with the `Authorize` ability, but grammatically, any ability is authorized in a sense. The possibility of `authorize` as a generic verb applicable to all abilities, overlapping with an Authorize ability, and being similar-but-related to access-checks (`Access::ReadWrite`) is a bit confusing. No concrete suggestions, still wrapping my head around all of it

rust/noosphere/src/ffi/authority.rs
```diff
@@ -0,0 +1,125 @@
...
+///
+/// The callback arguments are (in order):
+///
+///  1. The context argument provided in the original call to
+///     ns_sphere_authority_authorize
+///  2. An owned pointer to an ns_error_t if there was an error, otherwise NULL
+///  3. An owned pointer to constant string (the authorization CID) if the call
+///     was successful, otherwise NULL
+pub fn ns_sphere_authority_authorize(
```
jsantell, line 29: Requiring root authority explains some of my earlier questions I think, but some Q's:

* Why require root access? Wouldn't the primary/original author be able to delegate authority to another sphere, in terms of ucan?
* Certainly increases the necessity of storing the mnemonic on generation if primary author is not granted this permissions, and ultimately increases the necessity of good UX around this

rust/noosphere/src/ffi/petname.rs
```diff
@@ -84,7 +84,7 @@ pub fn ns_sphere_petname_get(
 /// The callback arguments are (in order):
 ///
 ///  1. The context argument provided in the original call to
-///     ns_sphere_content_read
+///     ns_sphere_petnames_assigned_Get
```
jsantell, line 87: nit: caps in function name

swift/Tests/SwiftNoosphereTests/NoosphereTests.swift
```diff
@@ -517,6 +517,70 @@ final class NoosphereTests: XCTestCase {
...
+        let sphere_mnemonic_ptr = ns_sphere_receipt_mnemonic(sphere_receipt, nil)
+        
+        let sphere = ns_sphere_open(noosphere, sphere_identity_ptr, nil)
+        
+        let expectation = self.expectation(description: "File contents are read")
+
+        nsSphereAuthorityAuthorize(noosphere, sphere, name: "alice", did: "did:key:alice") {
+            (error, authorization_ptr) in
+            
+            if error != nil {
```
jsantell, line 537: Do these error checks actually fail the test/return non-0 if `error != nil`?  edit: ah i see the `expectation` check

jsantell (commented):
Great work, and props for the comment updates! Some comments, mostly questions on how things work on my end

rust/noosphere-sphere/src/authority/escalate.rs
cdata: It's to encourage atomicity in the change being made with root authority. In the case of `SphereCursor`, it gives an opportunity to advance the cursor to a specific new version without the benefit of knowing what was done with the other `SphereContext`.

rust/noosphere/src/ffi/authority.rs
cdata: Root is currently only required for revocation; authorization can be done by anyone.

This is a shortcut. Revocation by a delegated key should be possible, but the UX is more complicated. For example: a delegatee cannot revoke authority from its delegator, and this nuance would have to be reflected in the UI on each device.

As we shore up the UX for delegation and revocation, we can begin to explore patterns that do not call for authority escalation.

rust/noosphere-core/src/authority/author.rs
```diff
@@ -71,24 +72,40 @@ where
         sphere_identity: &Did,
         db: &SphereDb<S>,
     ) -> Result<Access> {
+        let author_did = Did(self.key.get_did().await?);
+
+        // Check if this author _is_ the root sphere authority (e.g., when performing surgery on
+        // the authority section of a sphere)
+        if &author_did == sphere_identity {
```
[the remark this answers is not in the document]
cdata: A `SphereContext` always has an `Author`. In the case of escalating authority, `author.key` is derived from the mnemonic. `SphereAccess` is not a hard wall to enable or prevent writes, it is just used by `SphereContext` to throw an error if a caller tries to perform write operations in cases when we think that probably wouldn't pass muster if exposed to cryptographic scrutiny (in other words, it's a smoke test to help prevent footgun usage).

> the pattern used throughout is checking author == sphere, which always must imply using the Mnemonic/private key yeah?

Yes, but ultimately the test is a cryptographic signature check. Any signer that claims to be the sphere itself does not require a UCAN authorization, they just need the signature to be verifiable.

rust/noosphere-core/src/view/sphere.rs
```diff
@@ -791,6 +796,58 @@ impl<S: BlockStore> Sphere<S> {
...
+                    link
+                ));
+            }
+
+            if let Some(revocation) = revocations.get(&link).await? {
+                // NOTE: The implication here is that only the sphere itself can
+                // issue revocations The follow-on implicationis that a user
+                // must provide the sphere mnemonic in order to issue a
+                // revocation
+                if revocation.iss != sphere_identity {
```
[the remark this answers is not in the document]
cdata: Per the comment, the sphere root key always _issues_ the revocation. The sphere cannot revoke its own authority; it is the alpha and the omega, it is self-sovereign, what have you. And, as the root of all authority it may revoke any other authority delegated from it or another of its delegatees (direct or transitive).

PR #416, opened 2023-06-13
feat: re-enable checking of publishable timeframe when publishing link records from the gateway

### LINKED ISSUE #402: Validate ucan expiry when publishing from the gateway
Mechanisms were added in https://github.com/subconsciousnetwork/noosphere/pull/338 to test publishing expiry behind a flag. We'll need to enable this, as the lifetime will be short (~2 minutes) and the only place we validate the capability's lifetime.

jsantell:
Fixes #402


cdata (approved):
Woot

PR #419, opened 2023-06-13
fix: Update `vite` NPM package for security patch

cdata:
Fixes https://github.com/subconsciousnetwork/noosphere/security/dependabot/21


jsantell (approved):
:+1:

PR #437, opened 2023-06-21
Add support for sentry tracing in noosphere-core tracing

justinabrahms:


rust/noosphere-core/Cargo.toml
```diff
@@ -18,6 +18,10 @@ repository = "https://github.com/subconsciousnetwork/noosphere"
 homepage = "https://github.com/subconsciousnetwork/noosphere"
 readme = "README.md"
 
+[features]
+default = ["sentry"]
```
cdata, line 22: We should probably make it opt-in, not opt-out. Folks will only want it if they happen to have a Sentry account. And, if they do want it, it's easy to add a feature to their dependency configuration:

```toml
noosphere-core = { ..., features = ["sentry"] }
```

rust/noosphere-core/Cargo.toml
```diff
@@ -50,6 +54,8 @@ noosphere-collections = { version = "0.6.0", path = "../noosphere-collections" }
 
 ucan = { workspace = true }
 ucan-key-support = { workspace = true }
+sentry-tracing = { version = "0.31.5", optional = true }
+sentry = { version = "0.31.5", optional = true, default-features = false, features=["reqwest", "rustls", "backtrace", "contexts", "panic", "anyhow"] }
```
cdata, line 58: Did we verify if this is enough to evade the OpenSSL dependency?

rust/noosphere-core/src/tracing.rs
```diff
@@ -305,23 +309,45 @@ mod inner {
 
         let subscriber = tracing_subscriber::registry().with(env_filter);
 
+        // This isn't a secret value, just where to send the error reports. We can make this configurable in the future if other folks care.
+        let _guard = sentry::init(("https://553fd6eda33842ed9f088d0c16a147f1@o4505393671569408.ingest.sentry.io/4505399702126593", sentry::ClientOptions {
```
cdata, line 313: Is it specific to our account or is it a single endpoint used for all of Sentry? If it is specific to our account, then we should make it an environment variable.

cdata, line 313: We should only initialize Sentry if the feature flag is set (otherwise we'll get an error when it is not set, because the related dependencies won't be available).

rust/noosphere-core/src/tracing.rs
```diff
@@ -305,23 +309,45 @@ mod inner {
...
+            NoosphereLogFormat::Minimal => {
+                let subscriber = subscriber
+                    .with(
+                        Layer::default().event_format(NoosphereMinimalFormatter::new(
+                            tracing_subscriber::fmt::format()
+                                .without_time()
+                                .with_target(false)
+                                .with_ansi(USE_ANSI_COLORS),
+                        ))
+                    );
```
cdata, line 330: We should also include the `sentry_tracing::layer()` here.

cdata (changes requested):
Thanks so much for adding this :raised_hands: some light updates requested.

rust/noosphere-core/Cargo.toml
justinabrahms: yes.

rust/noosphere-core/src/tracing.rs
justinabrahms: Specific. Moved.

rust/noosphere-core/src/tracing.rs
```diff
@@ -305,23 +312,50 @@ mod inner {
 
         let subscriber = tracing_subscriber::registry().with(env_filter);
 
+        #[cfg(feature = "sentry")]
```
cdata, line 315: :100:

cdata (approved):
Great addition :beers: thanks for the contribution!

PR #438, opened 2023-06-22
Bump noosphere-core

justinabrahms:
This way, we can depend on the sentry changes elsewhere.


justinabrahms:
This happens from release-please instead.

PR #441, opened 2023-06-23
chore: Complain if a PR doesn't have a  release-please friendly title

justinabrahms:


.github/workflows/release-please-enforcer.yaml
```diff
@@ -0,0 +1,23 @@
+name: 'Release Please Enforcer'
```
cdata, line 1: Can we change the name of the workflow to something like lint-pull-request.yaml / "Lint PR"?

cdata (changes requested):

cdata (approved):
LGTM, let's clean up any issues after it lands

Issue #445, opened 2023-06-27
document how to use orb


justinabrahms:
so orb is a command line tool, but it turns out that you need to really understand noosphere's internals to make it work. When this issue is done, we should have documentation that shows how to create an identity, join a sphere, publish notes and read notes from others.

PR #447, opened 2023-06-27
chore: automatically render docs to the gh-pages branch

justinabrahms:


cdata:
Hey, thanks for opening this up for discussion.

I left a comment about a WIP docs branch in https://github.com/subconsciousnetwork/noosphere/issues/445

In [that branch](https://github.com/subconsciousnetwork/noosphere/tree/feat/noosphere-docs-website) we have a static docs site built with 11ty. The intention is for C docs, as well as Rust, TypeScript and Swift docs, to all be linked off of that site. In the mean time, I would probably say we should hold off publishing _just_ the C docs. We have a C FFI, but it's main point of consumption is via our Swft package, and I don't want to leave folks with the inaccurate impression that you need to be writing something in C/C++ just to use Noosphere.

justinabrahms:
@cdata the documentation in the repository suggests that the swift and typescript docs are not written. Are you thinking we should automate the building of our existing docs until someone spends more time writing docs? We could always just not tell people what the docs url is..

cdata:
> We could always just not tell people what the docs url is..

Good point. And, we should unblock this, since we will be generating these docs in CI here and it's really helpful that you're doing that.

Here's how I would like to proceed: let me suggest a change that generates the (as yet incomplete) docs site via CI (using a similar workflow to the one you've suggested for doxygen docs). Then, if you're game, you can rebase and we can get the doxygen docs integrated to the docs site.

cdata:
I've put up #456 for your consideration. After it lands, let's adapt this PR so that the doxygen docs can be generated and linked to from the 11ty-generated static site.

PR #448, opened 2023-06-27
chore: Documentation on how the CLI works

### LINKED ISSUE #445: document how to use orb

justinabrahms:
Fixes #445

### DIFF
--- a/rust/noosphere-cli/README.md
+++ b/rust/noosphere-cli/README.md
@@ -37,6 +37,7 @@ orb sync
 # sync data from a different sphere. Note, you'll need to
 # `orb auth add <did>` from that other sphere.
 orb sphere join <their DID> --local-key `whoami`
+orb sync # pull in their changes
 
 # then follow the onscreen instructions.
 ```
--- a/rust/noosphere-cli/README.md
+++ b/rust/noosphere-cli/README.md
@@ -28,7 +28,7 @@ orb status
 # persist changes to the sphere
 orb save
 
-# join a gateway, after you've set this identity as the counterpart
+# link your sphere to a gateway by setting this identity as the counterpart
 orb config set gateway-url <gatewayurl>
 
 # sync your changes with the upstream gateway
--- a/rust/noosphere-cli/README.md
+++ b/rust/noosphere-cli/README.md
@@ -5,3 +5,39 @@
 The Noosphere CLI is a reference client and pedagogical tool to demonstrate
 the principles of the Noosphere protocol and give interested users a
 no-code, low-complexity tool to synchronize content through the Noosphere.
+
+
+## Usage
+
+```sh
+# Create your identity
+orb key create `whoami`
+
+# make a directory for your sphere.
+mkdir my-sphere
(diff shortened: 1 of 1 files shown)


cdata:
LGTM, but need PR title to pass lint :)

rust/noosphere-cli/README.md
```diff
@@ -5,3 +5,56 @@
...
+
+# ..now make edits ..
+
+# see the status of files in your directory
+orb status
+
+# persist changes to the sphere
+orb save
+
+# join a gateway, after you've set this identity as the counterpart
```
jsantell, line 31: "join" means something else in this context e.g. in `orb sphere join`

jsantell, line 31: Wonder if we can just get the counterpart sphere from the gateway after setting, though makes things a bit magical. In lieu of that, maybe just have a line for that like `orb config set counterpart <counterpart_did>`

justinabrahms: How do you mean? From the linked video, this is what was typed to join the gateway.

cdata: Yah, this is room for refinement. Today it doesn't work that way, so the docs are accurate.

jsantell: i just mean the comment "join a gateway" in a pedantic-sense that maybe there's a better verb here (e.g. "register|subscribe to|set a gateway"), as we use `orb sphere join` later where "join" means synchronize another sphere (which is kinda what's going on here, admittedly)

rust/noosphere-cli/README.md
```diff
@@ -5,3 +5,56 @@
...
+orb config set gateway-url <gatewayurl>
+
+# sync your changes with the upstream gateway
+orb sync
+
+# sync data from a different sphere. Note, you'll need to
+# `orb auth add <did>` from that other sphere.
+orb sphere join <their DID> --local-key `whoami`
```
cdata, line 39: Sorry, just noticed this but: I think you need to join first, and then sync (currently steps as documented are in reverse order).

justinabrahms:
```suggestion
orb sphere join <their DID> --local-key `whoami`
orb sync # pull in their changes
```
It was intended as a different sync.

rust/noosphere-cli/README.md
```diff
@@ -5,3 +5,57 @@
...
+
+# ..now make edits ..
+
+# see the status of files in your directory
+orb status
+
+# persist changes to the sphere
+orb save
+
+# join a gateway, after you've set this identity as the counterpart
```
justinabrahms, line 31:
```suggestion
# link your sphere to a gateway by setting this identity as the counterpart
```

cdata (approved):
Thanks for documenting this!

PR #452, opened 2023-06-27
feat: 

justinabrahms:


cdata:
`orb status` actually does this already

justinabrahms:
It doesn't do it in such a way that is easily machine parsable. (I'm doing some scripting).

cdata:
Ah, that's a good point. This is some bike shedding on my part, but could you expose this as `orb status --id`, where the distinction is: "--id: only print the DID of the sphere"

My rationale at this time is:
- `orb config` currently pertains only to user-configurable key/value metadata that is local to the client (so getting the DID seems out of place there)
- `orb status` _does_ already display the DID, so a user may already think to look for this kind of flag there
- In other cases where we want to enable machine-friendly output, we use the same command + an `--as-json` flag (e.g., `orb key list --as-json`) and this is a little more in-line with that pattern

rust/noosphere-cli/src/native/mod.rs
```diff
@@ -178,6 +178,9 @@ pub enum ConfigGetCommand {
 
     /// Read the configured difftool command
     Difftool,
+
+    /// Get the identifier for the current sphere
+    DID,
```
jsantell, line 183: nit: prefer `Did` casing in rust outside of constants (this usually triggers a lint failure)

rust/noosphere-cli/src/native/mod.rs
```diff
@@ -96,7 +96,11 @@ pub enum OrbCommand {
 
     /// Show details about files in the sphere directory that have changed since
     /// the last time the sphere was saved
-    Status,
+    Status {
+        /// Output a single attribute of the status
+        #[clap(short, long, default_value="")]
+        attr: String,
```
cdata, line 102: If you make this field `id: bool` then it will automatically give you a `true` or `false` depending on if `--id` is used.

Then, the `status` command function can accept an `only_id` boolean.

cdata (changes requested):

rust/noosphere-cli/src/native/mod.rs
jsantell: Is there harm in outputting all the status info, rather than needing to specify a single attribute (IIUC)?

cdata: There is no harm, it's more about scope. The original change was pretty simple, but `orb status` is manually formatted and producing JSON output would be a significant increase in scope for this change. Ideally, we would have machine-friendly and human-friendly outputs for most commands.

justinabrahms: I'm not opposed to a json output.. it's just beyond my rust skills at the moment.

jsantell: For sure, `--json` or whatever is a bigger task for the future

PR #454, opened 2023-06-29
[DRAFT] UniFFI exploration

jsantell:
Exploring what it'd look like if we used [uniffi](https://github.com/mozilla/uniffi-rs) to generate our FFI bindings. Some challenges getting things set up with macro errors, but this is building and generating. Another challenge is to continue supporting C ABI/FFI as a baseline, and a separate object-oriented interface for dynamic languages via uniffi. We could approach this continuing our current pattern of FFI C functions interfacing with a uniffi-object (e.g. `NsNoosphere`), though this is just a very different way of designing APIs compared to e.g. `ns_error_code_get`

## Scaffolding

* All types used in the interface (and scaffolding macro) must be placed in `src/lib.rs`
* Even with `build.rs` pointing to udl file at `src/ffi/noosphere.udl`, bindgen still looks for udl at `src/noosphere.udl`.
* Ran into error during bindgen: "cargo metadata returned 0 packages" -- needed to change `rlib` crate type to `lib` https://github.com/mozilla/uniffi-rs/issues/1563

## Bindgen

* In lieu of using nightly, we have a `bin` to create bindings.
  * `cargo run --features=uniffi/cli --bin uniffi-bindgen generate --library ./target/debug/libnoosphere.so --language swift --out-dir ./target/ffi`
  * `--library` mode requires running in workspace root

## Generated examples:

Generates a `noosphere.h` (for uniffi internals, unrelated to our safer ffi bindings), swift bindings (and .modulemap) and kotlin:

### Kotlin

```kotlin
class NsNoosphere(
    pointer: Pointer
) : FFIObject(pointer), NsNoosphereInterface {
    constructor(`globalStoragePath`: String, `sphereStoragePath`: String, `gatewayApi`: Url?) :
        this(
    rustCallWithError(NoosphereException) { _status ->
    _UniFFILib.INSTANCE.uniffi_noosphere_fn_constructor_nsnoosphere_new(FfiConverterString.lower(`globalStoragePath`),FfiConverterString.lower(`sphereStoragePath`),FfiConverterOptionalTypeUrl.lower(`gatewayApi`),_status)
})

    /**
     * Disconnect the object from the underlying Rust object.
     *
     * It can be called more than once, but once called, interacting with the object
     * causes an `IllegalStateException`.
     *
     * Clients **must** call this method once done with the object, or cause a memory leak.
     */
    override protected fun freeRustArcPtr() {
        rustCall() { status ->
            _UniFFILib.INSTANCE.uniffi_noosphere_fn_free_nsnoosphere(this.pointer, status)
        }
    }
}
```

### Swift

```swift
public class NsNoosphere: NsNoosphereProtocol {
    fileprivate let pointer: UnsafeMutableRawPointer

    // TODO: We'd like this to be `private` but for Swifty reasons,
    // we can't implement `FfiConverter` without making this `required` and we can't
    // make it `required` without making it `public`.
    required init(unsafeFromRawPointer pointer: UnsafeMutableRawPointer) {
        self.pointer = pointer
    }
    public convenience init(`globalStoragePath`: String, `sphereStoragePath`: String, `gatewayApi`: Url?) throws {
        self.init(unsafeFromRawPointer: try rustCallWithError(FfiConverterTypeNoosphereError.lift) {
    uniffi_noosphere_fn_constructor_nsnoosphere_new(
        FfiConverterString.lower(`globalStoragePath`),
        FfiConverterString.lower(`sphereStoragePath`),
        FfiConverterOptionTypeUrl.lower(`gatewayApi`),$0)
})
    }

    deinit {
        try! rustCall { uniffi_noosphere_fn_free_nsnoosphere(pointer, $0) }
    }
}
```


jsantell:
[uniffi generated bindings for JS in firefox](https://hg.mozilla.org/mozilla-unified/rev/cc78aaf0481c92b16a48bc4b2cc2f2843cdbf234)

PR #470, opened 2023-07-06
chore: additionally publish images to ghcr.io

justinabrahms:
This should allow folks to use the images as well, and doesn't require us to have a magic google project id we have to remember.


cdata (approved):
Nice, thanks for adding this in!

PR #475, opened 2023-07-10
chore: fix header generation on macos

### LINKED ISSUE #473: Incomplete FFI header generation on macos
Running the FFI header generation on macos doesn't work locally (macos 13.2.1): a handful of headers generated by safer ffi doesn't actually output: `ns_sphere_petnames_assigned_get`, `ns_sphere_authority_escalate`, and a few other authority functions. We believe this is something with the generated output rather than these specific functions. 

Curiously enough, this works in [CI on macos 12](https://github.com/subconsciousnetwork/noosphere/blob/4940c331ed6fa0295b3ed93ec2f39133b3f973d2/.github/workflows/noosphere_apple_build.yaml#L87C8-L87C8).

As the swift framework bundles these headers, it's a confusing work around to run this locally. Some ideas:

* Prevent running header generation on macos. Better than thinking the header generated is valid.
* Investigate why this happens in safer ffi.
* Commit header to repo. Doesn't solve macos noosphere dev, though.

### LINKED ISSUE #476: C tests do not run on macos due to missing crypto libs
```
Undefined symbols for architecture x86_64:
  "_SecRandomCopyBytes", referenced from:
      ring::rand::darwin::fill::h4f46774401e5cd41 in libnoosphere.a(ring-470d6c82f7d73b3f.ring.a3b885f7-cgu.6.rcgu.o)
  "_kSecRandomDefault", referenced from:
      ring::rand::darwin::fill::h4f46774401e5cd41 in libnoosphere.a(ring-470d6c82f7d73b3f.ring.a3b885f7-cgu.6.rcgu.o)
ld: symbol(s) not found for architecture x86_64
```

jsantell:
Provide a script to generate headers that uses codegen_units=1 when building on macos to circumvent a header generation issue. Update CI and C makefile to use this script. Fixes #473.
Also fixes a linker issue on macos for the C example. Fixes #476.

### DIFF
--- a/scripts/build-framework.sh
+++ b/scripts/build-framework.sh
@@ -1,0 +1,101 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+SCRIPT_DIR=$(cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)
+PROJECT_DIR="$SCRIPT_DIR/../"
+cd $PROJECT_DIR
+
+OUT=""
+LITE=0
+RELEASE_FLAG=""
+PROFILE="debug"
+
+function usage() {
+  echo "build-framework.sh: a utility to build an Apple XCFramework for Noosphere."
+  echo "" 
+  echo "Usage:" 
+  echo "  build-framework.sh [options]" 
+  echo ""
+  echo "Options:"
+  echo "  -l --lite                  Only include x86_64-apple-darwin in framework."
+  echo "  -o <file>, --output <file> Output file."
+  echo "                             [default: \$ROOT/target/framework/{release,debug}/LibNoosphere.xcframework]"
+  echo "  --release                  Build artifacts in release-mode."
+  echo "  -h --help                  Show usage."
+  exit 0
+}
+
+while [[ $# -gt 0 ]]; do
+  case $1 in
+    -o|--output)
+      OUT="$2"
+      shift
+      shift
+      ;;
+    -l|--lite)
+      LITE=1
+      shift
+      ;;
+    --release)
+      RELEASE_FLAG="--release"
+      PROFILE="release"
+      shift
+      ;;
+    -h|--help)
+      usage
+      ;;
+    -*|--*)
+      echo "Unknown option $1"
+      exit 1
+      ;;
+  esac
+done
+
+if [[ ! -n ${OUT} ]]; then
+  OUT="$PROJECT_DIR/target/framework/$PROFILE/LibNoosphere.xcframework"
+fi
+
+rm -rf "$OUT"
+
+$SCRIPT_DIR/generate-headers.sh
+
+TARGETS=("x86_64-apple-darwin")
+if [[ $LITE -eq 0 ]]; then
+  TARGETS+=("aarch64-apple-ios")
+  TARGETS+=("x86_64-apple-ios")
+  TARGETS+=("aarch64-apple-ios-sim")
+  TARGETS+=("aarch64-apple-darwin")
+fi
+
+for TARGET in "${TARGETS[@]}"; do
+  rustup target install $TARGET
+  cargo build --package noosphere $RELEASE_FLAG --target $TARGET --locked
+done
+
+if [[ $LITE -eq 0 ]]; then
+  mkdir -p ./target/{macos-universal,simulator-universal}
+
+  lipo -create \
+    "./target/x86_64-apple-darwin/$PROFILE/libnoosphere.a" \
+    "./target/aarch64-apple-darwin/$PROFILE/libnoosphere.a" \
+    -output "./target/macos-universal/$PROFILE/libnoosphere.a"
+
+  lipo -create \
+    "./target/x86_64-apple-ios/$PROFILE/libnoosphere.a" \
+    "./target/aarch64-apple-ios-sim/$PROFILE/libnoosphere.a" \
+    -output "./target/simulator-universal/$PROFILE/libnoosphere.a"
+
+  xcodebuild -create-xcframework \
+    -library "./target/macos-universal/$PROFILE/libnoosphere.a" \
+    -headers ./target/headers/include/ \
+    -library "./target/simulator-universal/$PROFILE/libnoosphere.a" \
+    -headers ./target/headers/include/ \
+    -library "./target/aarch64-apple-ios/$PROFILE/libnoosphere.a" \
+    -headers ./target/headers/include/ \
+    -output "$OUT"
+else
+  xcodebuild -create-xcframework \
+    -library "./target/x86_64-apple-darwin/$PROFILE/libnoosphere.a" \
+    -headers ./target/headers/include/ \
+    -output "$OUT"
+fi
--- a/.github/workflows/noosphere_apple_build.yaml
+++ b/.github/workflows/noosphere_apple_build.yaml
@@ -96,13 +96,11 @@ jobs:
           brew: protobuf cmake
       - name: 'Generate the header'
         run: |
-          cd rust/noosphere/include/noosphere
-          cargo run --example generate_header --features headers --locked
-          cd -
+          ./scripts/generate-headers.sh
       - uses: actions/upload-artifact@v3
         with:
           name: include
-          path: rust/noosphere/include
+          path: target/headers/include
 
   # Generate Apple "universal" libraries for Noosphere for supported targets
   noosphere-apple-lipo:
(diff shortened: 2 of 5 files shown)


scripts/xcode_local_build
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
+set -e
+set -x
```
justinabrahms, line 3:
```suggestion
set -euxo pipefail
```

So you're missing `-u` which complains about unused variables (useful for dumb mistakes) and `-o pipefail` which says "if I use a pipe and it errors, then error".  Otherwise, you can have situations like:  

```sh
$ exit 1 | echo 'works'
works
```
when you really want it to error/exit.

scripts/xcode_local_build
```diff
@@ -0,0 +1,53 @@
...
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+PROJECT_DIR="$SCRIPT_DIR/../"
+
+pushd $PROJECT_DIR
```
justinabrahms, line 8:
```suggestion
cd $PROJECT_DIR
```

scripts/xcode_local_build
```diff
@@ -0,0 +1,53 @@
...
+)
+
+for TARGET in "${TARGETS[@]}"; do
+    rustup target install $TARGET
+    cargo build --package noosphere --release --target $TARGET --locked
+    # cargo build --package noosphert buie --target $TARGET --locked
+done
+
+mkdir -p ./target/macos
+mkdir -p ./target/simulator
```
justinabrahms, line 29:
```suggestion
mkdir -p ./target/{macos,simulator}
```

scripts/xcode_local_build
```diff
@@ -0,0 +1,53 @@
...
+    "x86_64-apple-ios"
+    "aarch64-apple-ios-sim"
+    "x86_64-apple-darwin"
+    "aarch64-apple-darwin"
+)
+
+for TARGET in "${TARGETS[@]}"; do
+    rustup target install $TARGET
+    cargo build --package noosphere --release --target $TARGET --locked
+    # cargo build --package noosphert buie --target $TARGET --locked
```
justinabrahms, line 25: I don't know what buie is, but maybe remove this?
```suggestion
```

scripts/xcode_local_build
```diff
@@ -0,0 +1,53 @@
...
+    -library ./target/simulator/libnoosphere.a \
+    -headers ./target/headers/include/ \
+    -library ./target/aarch64-apple-ios/release/libnoosphere.a \
+    -headers ./target/headers/include/ \
+    -output ./LibNoosphere.xcframework
+
+popd
+
+set +x
+set +e
```
justinabrahms, line 53:
```suggestion
```
You don't need to leave the directory or unset your flags. Bash scripts run as a subprocess and don't affect the outer scope. This would only be important if you were writing some library that gets sourced (which runs in your currents scope). This shouldn't be that.

justinabrahms (commented):
A few suggestions around bash scripts.

scripts/xcode_local_build
jsantell: ty for the feedback! I'll apply these throughout

jsantell:
Cleaned up the bash script from @justinabrahms' feedback; added a fix for #476 to link to `-framework Security` when running the C example on macos due to our `ring` dependencies.

justinabrahms (dismissed):

scripts/generate_headers
```diff
@@ -0,0 +1,21 @@
...
+cd $PROJECT_DIR
+
+rm -rf ./target/headers
+mkdir -p ./target/headers
+cp -r ./rust/noosphere/include ./target/headers/
+
+if [[ "$OSTYPE" == "darwin"* ]]; then
+	# macos linker fails to generate all exports without this flag
+	# https://github.com/subconsciousnetwork/noosphere/issues/473
+	CARGO_PROFILE_DEV_CODEGEN_UNITS=1 cargo run --verbose --package noosphere --example generate_header --features headers --locked
```
jsantell, line 16: @cdata this can be set via `[profile.dev]` for macos only, but don't believe we can target only the header gen example.

cdata: What are the side-effects? Is it just a perf difference?

jsantell: prevents LTO maybe? https://doc.rust-lang.org/cargo/reference/profiles.html#codegen-units

jsantell: serendipitous timing https://nnethercote.github.io/2023/07/11/back-end-parallelism-in-the-rust-compiler.html

cdata: Let's just stick to the script as written. We can revisit if it seems unsatisfying down the road, and I would rather not compromise all builds for the sake of this problem.

jsantell: Aye definitely not something we want on all builds, this just runs on macos's header-gen binary

cdata:
Conventionally we use `.sh` extensions for bash scripts. I don't think this is strictly necessary, but I would like to hear an argument for changing the convention before we commit to it.

I only have one argument in favor of `.sh`, which is that file extensions are valuable for inferring the file type without reading its contents (a topic top of mind for me as I work on `orb` file rendering).

jsantell:
Updated, adding `swift-test` to run swift tests locally, as well as adding some flags to manage debug/release builds. Consistent flag handling and usage docs added where needed.

These 3 scripts are consistent with themselves, but we need to make a few decisions on conventions; lets hammer out what those are and update here, or after, accordingly. What I'm using here are weak personal preferences:

* Indentation: 2 spaces
* Shebang: `#!/usr/bin/env bash`
* Path normalization:  Prefer `$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )` over `git rev-parse --show-toplevel`
* Naming: Use `-` (more bashy, easier to type) and omit extension (shebang clarifies any ambiguities if needed, will be executed more than read, and nicer when setting scripts in $PATH

scripts/generate_headers
jsantell: This gets the job done, but there's a side effect of needing to rebuild more often as this flag invalidates debug builds (that are missing this flag)

scripts/generate-headers
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)
+PROJECT_DIR="$SCRIPT_DIR/../"
+cd $PROJECT_DIR
+
+DEFAULT_OUT="`realpath $PROJECT_DIR`/target/headers"
```
jsantell, line 8: `realpath` needs `brew install coreutils` on Macos, and other solutions like `readlink -f` don't work on macos/BSD -- this is only really needed since it's displayed in the `usage`. do you have any cross-platform preferences for something like this? @justinabrahms

scripts/swift-test
```diff
@@ -0,0 +1,62 @@
...
+      ;;
+  esac
+done
+
+trap "mv $PLACEHOLDER Package.swift" EXIT
+
+cp Package.swift $PLACEHOLDER
+
+FRAMEWORK_PATH="./target/framework/$PROFILE/LibNoosphere.xcframework"
+echo $FRAMEWORK_PATH
```
jsantell, line 51: TODO: Remove

scripts/build-framework
```diff
@@ -0,0 +1,101 @@
...
+PROJECT_DIR="$SCRIPT_DIR/../"
+cd $PROJECT_DIR
+
+OUT=""
+LITE=0
+RELEASE_FLAG=""
+PROFILE="debug"
+
+function usage() {
+  echo "build-framework: a utility to build an Apple XCFramework for Noosphere."
```
jsantell, line 14: Just copying http://docopt.org/ styling here

scripts/build-framework.sh
```diff
@@ -0,0 +1,101 @@
...
+SCRIPT_DIR=$(cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)
+PROJECT_DIR="$SCRIPT_DIR/../"
+cd $PROJECT_DIR
+
+OUT=""
+LITE=0
+RELEASE_FLAG=""
+PROFILE="debug"
+
+function usage() {
```
cdata, line 13: :fire:

cdata (approved):
This is awesome, and clears up some huge FUD I had around header generation. Thanks @jsantell

PR #486, opened 2023-07-11
Open Telemetry

justinabrahms:
This assumes that #484 is done.

I'm interested in getting feedback about the code structure here before I start fighting with my lack of knowledge of rust. It seems inappropriate to have GCP specific code within noosphere. Instead, I'm thinking we should probably have a wrapper around noosphere which sets up opentelemetry, more cleanly separating the "record this metric" or "this log event happened" from what to do with it as a result.

Thoughts?

### DIFF
--- a/rust/noosphere-core/src/tracing.rs
+++ b/rust/noosphere-core/src/tracing.rs
@@ -2,6 +2,11 @@
 //! NOTE: [initialize_tracing] should only ever be called in tests or binaries;
 //! a library should only concern itself with instrumentation and logging.
 use strum_macros::{Display, EnumString};
+use opentelemetry::sdk::export::trace::stdout;
+use opentelemetry::runtime;
+use tracing_opentelemetry;
+use opentelemetry_gcloud_trace;
+use futures::executor;
 
 /// The crates that are considered when evaluating [NoosphereLog] and
 /// [NoosphereLogLevel] as directive configuration.
@@ -306,7 +311,15 @@ mod inner {
             env_filter = env_filter.add_directive(directive)
         }
 
-        let subscriber = tracing_subscriber::registry().with(env_filter);
+        let tracer = futures::executor::block_on(opentelemetry_gcloud_trace::GcpCloudTraceExporterBuilder::new("GCP ID @@@".to_string())
+            .install_batch(opentelemetry::runtime::Tokio));
+
+        let otel = tracing_opentelemetry::layer().with_tracer(tracer);
+        let subscriber = tracing_subscriber::registry()
+            .with(env_filter)
+            .with(otel)
+            .try_init()?;
+        let subscriber = subscriber.with(otel);
 
         match noosphere_log_format {
             NoosphereLogFormat::Minimal => {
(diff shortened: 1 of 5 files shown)


rust/noosphere-core/src/tracing.rs
```diff
@@ -337,6 +353,15 @@ mod inner {
...
                 let subscriber = subscriber.with(sentry_tracing::layer());
 
+                subscriber.init();
+            }
+            NoosphereLogFormat::Structured => {
+                let subscriber =
+                    subscriber.with(Layer::default().json().with_ansi(USE_ANSI_COLORS));
+
+                #[cfg(feature = "sentry")]
+                let subscriber = subscriber.with(sentry_tracing::layer());
```
jsantell, line 363: Could we take a similar approach with Open Telemetry as Sentry (make dependencies optional, put it behind a feature flag, require building with a GCP ID etc.)?

justinabrahms: We could, yeah, but it just seems inappropriate to have in the repo entirely?

Tracing / OTel / etc separate out their exporters from their initial collection so that open source projects can instrument things, but not be concerned about how that actually gets to the place you want. Having to bake in "oh, we use sentry" or "oh, we use GCP" into the open source code feels off to me. If anything, I think we should move this tracing initialization code into a private repository makes sense if we want to share configuration (which seems like a good idea?) but I don't think we should burden the broader open source community with our particular operational needs.

jsantell: :100:

cdata: It does seem like "bring your own tracing subscriber" is a pattern we should be thinking more carefully about. That said: the only case where a user can take advantage of such capability is a Rust author. Today we also serve TypeScript and Swift authors (and arguably C), and these users will depend on the ability to consume an artifact that contains the functionality they need.

We could theoretically offer a FFI that enables custom subscriber layers whose behavior is defined on the other side of FFI (this came up when discussing potential improvements to logging with @bfollington ). But, that would be a more involved change.

The fastest path is arguably to integrate OpenTelemetry support via conditional compilation. And, it has the appeal of enabling us to produce artifacts for other language environments that have the desired capabilities built-in (though it comes with the trade-off that it requires multiple configuration points to use this way).

As far as separation of concerns goes: we definitely shouldn't leak "we use GCP" into this layer. At the minimum, I think that the "tracer" should be provided as optional configuration (and some FFI can perhaps be offered to configure the tracer).

My conclusion at this point is that we need to review what deploying open telemetry looks like on the Subconscious side; it may be the case that we _have_ to initialize it once within the app in order to use it correctly, in which case we are forced to reckon with the FFI boundary.

justinabrahms: I've kicked off the FFI question with the CNCF's tag-observability. I think there are benefits to landing this even without FFI (gateway & cloudctl traces).

cdata: Yah, that's fine, but it is worthwhile to speculate about the FFI implications because it may weigh on the shape of the solution here.

Perhaps the least speculative approach would be to have a feature flag that enables otel, and changes the signature of `initialize_tracing` to accept a second parameter that is `Option<Tracer>`. Then, we can configure otel in a cloud-agnostic way.

PR #492, opened 2023-07-12
chore: Replace noosphere-car with iroh-car throughout the noosphere crates.

### LINKED ISSUE #446: Deprecate `noosphere-car`, switch back to `iroh-car`
@b5 and @dignifiedquire have agreed to upstream our patches to `iroh-car` and begin maintaining the crate again. So, although we had lots of good times together, it's probably time to part ways with `noosphere-car`. The PR where they upstream our changes can be found here: https://github.com/n0-computer/iroh-car/pull/6

As our tweaks are upstreamed and land in a crates.io release, we can begin work to swap `noosphere-car` out for `iroh-car`.

jsantell:
Fixes #446

TBD what we have to do to mark noosphere-car as deprecated -- will need to remove the code and workspace/release-it entries

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2451,6 +2451,21 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "12b6ee2129af8d4fb011108c73d99a1b83a85977f23b82460c0ae2e25bb4b57f"
 
 [[package]]
+name = "iroh-car"
+version = "0.3.0"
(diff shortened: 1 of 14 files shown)


cdata:
Awesome! Can you file a companion issue to deprecate `noosphere-car` so that we don't lose track of it?

jsantell:
Added `#[deprecated]` to exported functions, module docs, and README. I think after a release cycle, we can remove it from our workspace.

PR #500, opened 2023-07-15
feat: CLI local-key -> pet-name

justinabrahms:


justinabrahms:
Just realizing that this is literally the my own sphere's key.

PR #514, opened 2023-07-19
fix: Get `chromedriver` from a new location

cdata:
Our workflow script for getting `chromedriver` for our browser-targeted tests broke from under us due to changes in the Chromium project's release processes (see: https://groups.google.com/g/chromedriver-users/c/clpipqvOGjE).

We originally cribbed the script from https://github.com/SeleniumHQ/selenium/blob/5d108f9a679634af0bbc387e7e3811bc1565912b/.github/actions/setup-chrome/action.yml

However, Selenium has moved to using https://github.com/browser-actions/setup-chrome which does not handle setting up `chromedriver`.

The original mechanism we depended on to discover the correct `chromedriver` version has been replaced by the static JSON published in https://github.com/GoogleChromeLabs/chrome-for-testing#json-api-endpoints

So, this change adapts our workflow scripts to use the new mechanism to setup `chromedriver`.

### DIFF
--- a/.github/workflows/run_test_suite.yaml
+++ b/.github/workflows/run_test_suite.yaml
@@ -43,7 +43,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: Swatinem/rust-cache@v2
       - name: 'Setup Rust'
-        run: | 
+        run: |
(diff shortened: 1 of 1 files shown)


jsantell (approved):
LGTM!

PR #521, opened 2023-07-21
feat: Safer cleanup when dropping Noosphere

cdata:
This change proposes the following:

- A blocking cleanup step to allow async work scheduled by Tokio to finish when invoking `ns_free`
- A scaffold for fuzzing the C FFI from the perspective of Swift
- Fixes some already-shipped FFI so that they are actually available to Swift
- Bumps our MSRV to 1.70 (because of `std::cell::OnceCell` use in C FFI)


rust/noosphere/src/sphere/builder.rs
```diff
@@ -30,18 +30,14 @@ use crate::{
     storage::StorageLayout,
 };
 
+#[derive(Default)]
 enum SphereInitialization {
+    #[default]
```
jsantell, line 35: TIL

jsantell (approved):
Nice fuzzing work :+1:

PR #530, opened 2023-07-26
feat!: `orb` uses latest Noosphere capabilities

cdata:
This change proposes a significant re-vamp / update for the Noosphere CLI (known as `orb`). The CLI has limped along and kept "working" for some time since we first implemented it last Fall. However, it has not received updates to make use of the latest Noosphere capabilities (such as petnames). And, there were some bugs here and there lingering in the uncovered surface area of the implementation.

- New `auth` subcommand to manage sphere authority
- New `follow` subcommand to follow or unfollow other spheres
- Spheres render peers in address book to filesystem
- New integration tests specifically targetting CLI
- Removes several unimplemented commands
- Many sub-commands have been regrouped underneath `orb sphere`


cdata:
Related to `safer-ffi` feature flag change: https://github.com/getditto/safer_ffi/issues/175

layout.md
```diff
@@ -0,0 +1,78 @@
+```sh
+/sphere_root/
+ ├── .sphere/
+ │   ├── identity
+ │   ├── version  # Last rendered version of the root sphere
+ │   ├── slugs    # Backward mapping of slugs; flat directory where slugs have '∕' in place of '/'
+ │   │   ├── foo -> ../../foo.subtext
+ │   │   └── bar∕baz -> ../../bar/baz.subtext
+ │   ├── storage/ # Storage folder distinguishes the root sphere
```
justinabrahms, line 9: So this also stores the content for the peers?

rust/noosphere-cli/src/native/commands/auth.rs
```diff
@@ -120,62 +72,184 @@ IMPORTANT: You MUST sync to enable your gateway to recognize the authorization:
...
+                    "hierarchy": hierarchy,
+                    "meta": meta
+                }))?
+            );
+        } else {
+            info!(" ⌀ Sphere");
+            info!(" │ {sphere_identity}");
+            info!(" │");
+
+            fn draw_branch(
```
justinabrahms, line 185: Rust newbie question:  In other programming languages, i would have said the following. I'm not certain if it applies. Thoughts?


Generating a big method like this inside of another method seems odd. Should this live elsewhere?
Also, you have differing functionality using the same input (json vs ascii tree rendering). Is there a better programming construct to separate these, like an interface/trait?

rust/noosphere-cli/src/native/commands/auth.rs
```diff
@@ -120,62 +72,184 @@ IMPORTANT: You MUST sync to enable your gateway to recognize the authorization:
...
+                    .collect::<String>();
+
+                while let Some(link) = items.pop() {
+                    let is_last = items.is_empty();
+
+                    if let Some((name, id, _token)) = meta.get(&link) {
+                        let (branch_char, trunk_char) = match is_last {
+                            true => ('└', ' '),
+                            false => ('├', '│'),
+                        };
```
justinabrahms, line 209: 😍

layout.md
cdata: Yup

rust/noosphere-cli/src/native/commands/auth.rs
cdata: This is pretty fresh (written today), I wouldn't consider it a stylistic reference.

Similar response to "differing functionality" in this case.

As for factoring, there are two questions I would ask myself:

- Is there obvious headroom for generalization across branches?
- How frequently will I benefit from a generic interface?

Rust gives us traits, which are similar to (but not quite the same as) interfaces. Traits can be a powerful tool for defining generic behavior, but they come with non-zero cognitive and syntactical overhead in order to use them.

In this case, the two render modes are pretty different in nature (one uses `serde_json`, the other is all manual formatting). And, most branches are pretty small. Pulling out different branches into their own functions is a relatively easy, low-overhead way to clean it up, and you can even use the signature as an approximation for a generic interface.

If we reached a point where rendering was made to be pluggable, or at least had more than a small handful of modes, I might consider a more sophisticated / extensible approach using traits and generics.

After all, it's always nice to have the headroom to refactor into baroque abstractions down the road :sweat_smile:

layout.md
cdata: I should mention the contents of `layout.md` are my personal notes, not documentation, and the structure depicted is not representative of the actual implementation.

justinabrahms: Here's the note to please add it to some form of real documentation b/c it seems really really helpful. :D

rust/noosphere-cli/src/native/commands/auth.rs
justinabrahms: Another version of this might be: Is there benefit to making the Vec that contains these a Real Type™ and then putting those serialization options on it? My understanding is that rustaceans tend to prefer NOT to do this sort of thing.

cdata: > Another version of this might be: Is there benefit to making the Vec that contains these a Real Type™ and then putting those serialization options on it? My understanding is that rustaceans tend to prefer NOT to do this sort of thing.

Definitely! That can be a nice way to encapsulate behavior. I would probably come down on: it's a stylistic choice in this case. But, preferences aside, I think it's a valid way to express things like this. If there is an obvious advantage to doing it that way compared to a few small functions, go for it!

To speculate, I imagine that one case where Rust programmers may avoid using a stateful struct is when they are dealing in a lot of shared data and want to keep overhead as low as possible (both resource- and cognition-wise). In this case, one likely needs to perform acrobatics with lifetimes; these can be harder to express when you have structs keeping references to shared data. The lifetimes usually can be expressed, but the effort to do it may serve as a friction. It is sometimes easier to express these cases as stateless functions.

justinabrahms:
Tried this out locally on an existing sphere I have. How do I show it where my key is?

```
 ✘  abrahms@subconscious-nix  ~/noosphere/sphere  orb sphere auth list
Error: No key named mykey found!
 ✘  abrahms@subconscious-nix  ~/noosphere/sphere  orb sphere status   
Error: No key named mykey found!
 abrahms@subconscious-nix  ~/noosphere/sphere  find ~/.noosphere/     
/home/abrahms/.noosphere/
/home/abrahms/.noosphere/keys
/home/abrahms/.noosphere/keys/mykey.public
/home/abrahms/.noosphere/keys/mykey.private
/home/abrahms/.noosphere/keys/notesync.public
/home/abrahms/.noosphere/keys/notesync.private

```

cdata:
Thanks for the callout about the global `.noosphere` directory @justinabrahms 

As of this change, we will be dynamically selecting the default global configuration directory using https://crates.io/crates/directories, so the old location (which may or may not have even been valid in non-WSL Windows) is no longer being used anywhere.

As before, the location can still be customized by embedders of the Noosphere library.

rust/noosphere-cli/src/native/render/renderer.rs
```diff
@@ -0,0 +1,264 @@
...
+/// [SphereRenderer] embodies all of the work needed to render a sphere graph to
+/// a workspace file system location. Starting from the "root" sphere, the
+/// renderer will efficiently queue work to concurrently render the sphere graph
+/// up to a maximum depth.
+///
+/// The renderer produces a file system structure that is approximated by this
+/// diagram:
+///
+/// ```sh
+/// /workspace_root/
```
jsantell, line 25: :fire: :fire:

rust/noosphere-cli/src/native/render/renderer.rs
```diff
@@ -0,0 +1,264 @@
...
+                            let SphereRenderRequest(petname_path, peer, version, link_record) = job_request;
+
+                            // NOTE: If a peer is too deep, we _don't_ mark it as started; another
+                            // peer may wish to render this peer at a shallower depth, in which case
+                            // we should proceed.
+                            if petname_path.len() > render_depth as usize {
+                                debug!("Skipping render job for '@{}' (exceeds max render depth {render_depth})", petname_path.join("."));
+                                continue;
+                            }
+                            warn!("PETNAME PATH {:?}", petname_path);
```
jsantell, line 215: nit: remove?

rust/noosphere-cli/tests/peer_to_peer.rs
```diff
@@ -552,3 +563,110 @@ async fn replicate_older_version_of_peer_than_the_one_you_have() -> Result<()> {
...
+
+    let mut pair_1 = SpherePair::new("pair_1", &ipfs_url, &ns_url).await?;
+    let mut pair_2 = SpherePair::new("pair_2", &ipfs_url, &ns_url).await?;
+
+    pair_1.start_gateway().await?;
+    pair_2.start_gateway().await?;
+
+    pair_2
+        .spawn(move |mut ctx| async move {
+            // ctx.write(slug, content_type, value, additional_headers)
```
jsantell, line 582: nit: remove

rust/noosphere-sphere/src/authority/read.rs
```diff
@@ -30,9 +30,12 @@ where
         authorization: &Authorization,
     ) -> Result<(), NoosphereError>;
 
-    /// Look up an authorization by a [Did].
+    /// Look up an [Authorization] by a [Did].
     async fn get_authorization(&self, did: &Did) -> Result<Option<Authorization>>;
 
+    /// Look up all [Authorization]s with the specified name
```
jsantell, line 36: Consider: device name/key name

jsantell (approved):
Great work!

rust/noosphere-cli/src/native/commands/sphere/follow.rs
```diff
@@ -0,0 +1,297 @@
...
+use crate::native::{commands::sphere::save, workspace::Workspace};
+use anyhow::{anyhow, Result};
+use noosphere_core::data::Did;
+use noosphere_sphere::{
+    HasMutableSphereContext, SpherePetnameRead, SpherePetnameWrite, SphereWalker,
+};
+use serde_json::{json, Value};
+use tokio_stream::StreamExt;
+
+/// Add a peer to your address book by assigning their sphere ID to a nickname
```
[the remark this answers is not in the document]
cdata: Filed this for further discussion https://github.com/subconsciousnetwork/noosphere/issues/563

PR #535, opened 2023-07-27
fix: Allow expired LinkRecords to be republished by the gateway. Fixes #533

### LINKED ISSUE #533: `LinkRecord` republishing in gateway is broken due to expiry checks
In https://github.com/subconsciousnetwork/noosphere/pull/416 we re-enabled expiry checks for link records. I think this had the unintended side-effect of cancelling out the gateway's ability to re-publish link records periodically (because every time it attempts to republish, the record is seen to be expired and so is discarded). We should reconcile this so that we can preserve gateway re-publishing of link records.

jsantell:


cdata (approved):
<3 thanks for the turnaround on this

PR #539, opened 2023-07-28
chore: mark noosphere-car's version as 0.3.0+deprecated

jsantell:
* Structs needed to be marked as deprecated in order to float the deprecation to the docs landing page (the export was insufficient)
* Should be sure that the release process doesn't increment or change the version on publish (though if it bumps to `0.3.1+deprecated`, oh well)
* FWIW `0.3.0` < `0.3.0+deprecated`

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -7,7 +7,9 @@
   "last-release-sha": "39d0260c3fd2e4deb155f7eb60d460c0d76dae4f",
   "packages": {
     "rust/noosphere-api": {},
-    "rust/noosphere-car": {},
+    "rust/noosphere-car": {
+      "release-as": "0.3.0+deprecated" 
+    },
     "rust/noosphere-cli": {
       "draft": true
     },
--- a/rust/noosphere-car/src/header.rs
+++ b/rust/noosphere-car/src/header.rs
@@ -7,6 +7,7 @@ use crate::error::Error;
 /// A car header.
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[non_exhaustive]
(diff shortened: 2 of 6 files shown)


jsantell:
Using a deprecated package fails lints :monocle_face:  :upside_down_face: 

```
error: use of deprecated field `writer::CarWriter::writer`: `noosphere-car` is deprecated. Please use `iroh-car`.
  --> rust/noosphere-car/src/writer.rs:75:9
   |
75 |         self.writer
   |         ^^^^^^^^^^^
```

cdata:
That's confusing; is the error occurring within the package itself?

cdata:
Otherwise: are we fully switched to `iroh-car` everywhere else?

jsantell:
> That's confusing; is the error occurring within the package itself?

Yes! `cargo clippy --all -- -D warnings` in `noosphere-car`

> Otherwise: are we fully switched to `iroh-car` everywhere else?

Yes!

cdata:
@jsantell some suggestions so that we can land this change:

- Let's disable the failing lint within the `noosphere-car` package via `#[allow(...)]` macro. IIUC that should allow the lint to pass, and won't affect other packages since we no longer depend on this one within our workspace.
- Don't change the version directly; instead, we should use the `release-please` configuration to specify a `release-as` version. For example: https://github.com/subconsciousnetwork/noosphere/blob/3818134845007581dfc770eef3380cf16d8b59d4/release-please-config.json#L8-L10

jsantell:
* updated permissive linting within noosphere-car
* updated version with release please config

cdata (approved):
Woot!

PR #540, opened 2023-07-28
feat: Synchronously flush sled DB

### LINKED ISSUE #403: sled::Tree::flush_async() deadlocks: Intermittent test failures
I've seen this happen occasionally when running tests locally.

```
test single_player_single_device_end_to_end_workflow has been running for over 60 seconds
test writes_do_not_block_reads has been running for over 60 seconds
```

jsantell:
.. rather than using sled::Tree::flush_async, which can cause deadlocks. Fixes #403

### DIFF
--- a/rust/noosphere-storage/src/implementation/native.rs
+++ b/rust/noosphere-storage/src/implementation/native.rs
@@ -85,7 +85,11 @@ impl Store for NativeStore {
 
     /// Flushes pending writes if there are any
     async fn flush(&self) -> Result<()> {
-        self.db.flush_async().await?;
+        // `flush_async()` can deadlock when simultaneous calls are performed.
+        // This occurs often in tests and fixed in `sled`'s main branch,
+        // but no cargo release since 2021.
+        // https://github.com/spacejam/sled/issues/1308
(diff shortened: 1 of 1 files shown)


rust/noosphere-storage/src/implementation/native.rs
```diff
@@ -85,7 +85,12 @@ impl Store for NativeStore {
 
     /// Flushes pending writes if there are any
     async fn flush(&self) -> Result<()> {
-        self.db.flush_async().await?;
-        Ok(())
+        // `flush_async()` deadlocks until there's a new release
+        // of sled. In the mean time, wrap it in our own async.
+        // https://github.com/spacejam/sled/issues/1308
+        let db = self.db.clone();
+        tokio::spawn(async move { db.flush() })
```
cdata, line 92: From a scheduling perspective, I'm not sure if it makes that much of a difference to wrap this with `tokio::spawn`. Maybe we should just do a blocking flush for now?

cdata (approved):

cdata (approved):
Woot!

PR #546, opened 2023-08-01
chore: promote url crate to workspace dependency.

jsantell:


Cargo.toml
```diff
@@ -40,6 +40,7 @@ tracing = { version = "0.1" }
 tracing-subscriber = { version = "~0.3.16", features = ["env-filter", "tracing-log", "json"] }
 ucan = { version = "0.4.0" }
 ucan-key-support = { version = "0.1.7" }
+url = { version = "^2.4" }
```
cdata, line 43: Do we require this level of specificity in the dependency? If not, I would suggest keeping the range open to any 2.x.x version.

jsantell: sg!

cdata (approved):

PR #548, opened 2023-08-02
Add version for orb command

justinabrahms:
```
$ orb --version
orb fd14912-modified
```

^^ locally w/ this change, pre-commit

For the release build, this should result in the tag. It will be set to the value of `git describe --always --dirty=-modified`.

Unfortunately, github (and therefore release-please) doesn't make annotated tags on a release. https://github.com/googleapis/release-please/issues/2034

This means that we either needs some custom logic around moving our release-please generated tags to annotated tags.. or I can write more additional code to alter the display logic to not be based on git.

Making this a draft to figure out the preferred direction. Thoughts, @cdata ?


cdata:
This sounds like a good feature to have. I'm actually unfamiliar with the difference between light and annotated tags - can you share some details on those?

justinabrahms:
From `man git-tag`:
> Tag objects (created with -a, -s, or -u) are called "annotated" tags; they contain a creation date, the tagger name and e-mail, a tagging message, and an optional GnuPG signature. Whereas a "lightweight" tag is simply a name for an object (usually a commit object).

justinabrahms:
Closing this. Will re-open when we get a release-please alternative that does annotated tagging.

PR #551, opened 2023-08-03
fix: Ensure write access when syncing a sphere. Fixes #389

### LINKED ISSUE #389: Disallow sync on read-only sphere contexts
Currently we check most write operations to verify that the author is read-write access to the sphere context before proceeding. Although there is no danger that the author can create valid sphere history when these writes are allowed, the check enables us to surface an appropriate error / warning to the writer that they attempted a write that would not be allowed.

We do not currently do this change on sync, even though sync may change the local sphere history (effectively a write). This has come up at least once when someone tried to sync a sphere that was traversed to (so definitely was read-only) and got an obtuse error from Noosphere.

So, we should perform this check for read-write access before proceeding with a sync.

jsantell:


rust/noosphere-sphere/src/sync/error.rs
```diff
@@ -5,6 +5,9 @@ use thiserror::Error;
 /// gateway
 #[derive(Error, Debug)]
 pub enum SyncError {
+    /// The error was due to not having write access to the sphere
+    #[error("Insufficient permission to sync")]
+    NoPermission,
```
cdata, line 10: My bikeshed here is: name it `InsufficientPermission` :bike:

cdata:
I poked at this a little and I think we need to do the following: special "allowed" case for when we are syncing, we have an `Authorization::Cid` and we don't have any history for the sphere at all. Maybe we could do this by a special parameter passed to the call to `sync`? E.g., `sync(first_time: bool)` or something...... IDK that's kinda cludgey, but WDYT? 

### Rationale

There is nothing strictly wrong with trying to sync when you don't have read/write access; we just want to help the user avoid the footgun of _trying_ to sync a sphere that they don't have write access to (because it is almost guaranteed to fail except in the one-time "join" case).

And, at the same time, there is a case where we have will read-only access BUT we should attempt to sync anyway: when joining a sphere for the first time. In this condition, we have a CID for an authorization but no sphere history (nor the authorization itself), and we are saying to the gateway: I have a CID for an authorization that I expect you to have, please let me sync _any_ history so that I can get my authorization as well as the sphere content. From there, it is in the gateway's hands to decide if the authorization actually exists or not.

jsantell:
Updating this to check if the sphere's author is `Authorization::Ucan`, ensure write access is available, otherwise (`Authorization::Cid`) rely on the gateway. For completeness with the gateway being the final arbiter, we could propagate the auth checks from the gateway/`noosphere-api` to sphere client/`noosphere-sphere` types for downstream handling as well

jsantell:
Updated from the `noosphere-sphere` removal, but now with recovery implemented, we also need to special case that scenario too.

PR #578, opened 2023-08-08
feat: Ensure adopted link records are fresher than previous entries

### LINKED ISSUE #258: Ensure that name records are the most up to date during adoption
As of #253 , we will have end-to-end name publishing and resolving. However, we don't currently verify that a given record that we are adopting is the most up to date record we have adopted for a given entry in our address book. This verification can (and probably should) be done by comparing the lifetime constraints of the name records in question, and selecting the record with the most recent "not before" (`nbf`) value.

### LINKED ISSUE #260: Determine policy when name record can no longer be resolved for a DID
In some cases, we will try to resolve an updated record for a DID that we have previously resolved a record for, only to find that we are no longer able to resolve any record for that DID. We should determine a policy for how to persist (or expire) records when no record is able to be resolved for the corresponding DID.

### LINKED ISSUE #562: Compare expiry when considering whether to adopt link records
When considering whether or not to adopt a new link record for a peer where a valid link record is already available and the link records refer to the same link, we should compare expiry bounds to determine which of the link records is "fresher," and prefer that one.

jsantell:
Fixes #258, fixes #562

### DIFF
--- a/rust/noosphere-sphere/src/petname/write.rs
+++ b/rust/noosphere-sphere/src/petname/write.rs
@@ -137,6 +137,14 @@ where
             return Err(anyhow!("Sphere cannot assign itself to a petname."));
         }
 
+        if let Some(existing_record) = self.get_petname_record(name).await? {
+            if !existing_record.superceded_by(record) {
+                return Err(anyhow!(
+                    "Previously stored record supercedes provided record."
+                ));
+            }
+        }
+
         let cid = self
             .sphere_context_mut()
             .await?
@@ -144,7 +152,6 @@ where
             .write_token(&record.encode()?)
             .await?;
 
-        // TODO: Verify that a record for an existing address is actually newer than the old one
         // TODO: Validate the record as a UCAN
 
         debug!(
--- a/rust/noosphere-core/src/data/address.rs
+++ b/rust/noosphere-core/src/data/address.rs
@@ -170,6 +170,21 @@ impl LinkRecord {
         None
     }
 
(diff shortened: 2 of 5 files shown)


rust/noosphere-sphere/src/petname/write.rs
```diff
@@ -137,14 +137,21 @@ where
             return Err(anyhow!("Sphere cannot assign itself to a petname."));
         }
 
+        if let Some(existing_record) = self.get_petname_record(name).await? {
+            if !existing_record.superceded_by(record) {
+                return Err(anyhow!(
```
cdata, line 142: We probably want to make sure that this isn't causing failures where we previously wouldn't have had any during sync: https://github.com/subconsciousnetwork/noosphere/blob/a6f0a74cde2fc001bfff5c1bed0844ac19fc8258/rust/noosphere-sphere/src/sync/strategy.rs#L314

This kind of validation error probably needs to be recovered-from in most cases where it happens. Most likely we have other name records for consideration and one invalid record should not blow up the flow if the others could still be used.

cdata:
Great work on this, I think it needs some consideration in the periphery of usage but otherwise I like it.

rust/noosphere-sphere/src/petname/write.rs
jsantell: Good catch!

jsantell:
Updated with handling the error in non-test/looping scenarios where we want to continue (:drum:) processing if a failure occurs

cdata (approved):

Issue #579, opened 2023-08-08
Intermittent Windows file lock error


jsantell:
Intermittently, a locking error shows up on Windows CI builds: "The process cannot access the file because another process has locked a portion of the file."

[Windows Run](https://github.com/subconsciousnetwork/noosphere/actions/runs/5787862436/job/15685696356?pr=538)
```
thread 'gateway_receives_a_newly_initialized_sphere_from_the_client' panicked at 'called `Result::unwrap()`
on an `Err` value: IO error: could not acquire lock on
 "C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\.tmpjw4OSJ\\.sphere/storage\\db":
Os { code: 33, kind: Uncategorized, message: "The process cannot access the file because another process
has locked a portion of the file." }', rust\noosphere-cli\tests\gateway.rs:185:77
```

[Windows Run](https://github.com/subconsciousnetwork/noosphere/actions/runs/5800800263/job/15723665892?pr=538)

```
thread 'single_player_single_device_end_to_end_workflow' panicked at 'called `Result::unwrap()`
 on an `Err` value: IO error: could not acquire lock on
 "C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\.tmpj2ZHt3\\did_key_z6MkvfZ1aRD4R6wiqDaQqP6NQdXyc4vXfA9ScfSq8Yu7eEpf\\db":
 Os { code: 33, kind: Uncategorized, message: "The process cannot access the file because another process has locked a portion of the file." }', rust\noosphere\tests\integration.rs:143:14
```

jsantell:
This could possibly be due to closing and reopening Sled, [considered unsupported behaviour](https://github.com/spacejam/sled/issues/1234). In the tests this occurs in, we clearly create a sphere and its storage layer, close it, and try to reopen.

* single_player_single_device_end_to_end_workflow
* sphere::builder::tests::it_can_create_a_scoped_sphere_and_later_open_it
* gateway_receives_a_newly_initialized_sphere_from_the_client
  * We `sphere_create(..)` followed by the Workspace initializing a sphere context

jsantell:
via https://github.com/subconsciousnetwork/noosphere/issues/167, the same lock error in `sphere::builder::tests:it_can_create_a_sphere_and_later_open_it`

> ---- sphere::builder::tests::it_can_create_a_sphere_and_later_open_it stdout ----
[...]

PR #580, opened 2023-08-08
feat: flush when dropping NativeStorage to prevent intermittent failures

jsantell:


cdata (approved):
Nice!

jsantell:
MAY fix #579, and #167

Issue #498, opened 2023-07-15
Potentially unbounded memory overhead using `BodyChunkIpld`


cdata:
Basic usage of one of our most venerable constructs, `BodyChunkIpld`, will lead to unacceptable levels of memory overhead when reading large files. Currently, both `BodyChunkIpld::store_bytes` and `BodyChunkIpld::load_all_bytes` allocate up to file-size memory when used.

In the case of `BodyChunkIpld::store_bytes`: in order to _efficiently_ encode/store bytes from a file as `BodyChunkIpld` pages, we would need to read the file back-to-front. This introduces some complexity to how we compute chunk cut points (using a Rust implementation of [FastCDC](https://www.usenix.org/system/files/conference/atc16/atc16-paper-xia.pdf)). [We contributed a change to our dependency to enable async streaming support in the chunker](https://github.com/nlfiedler/fastcdc-rs/pull/32), which gets us half-way to where we need to be. However, in order to make use it we will most likely have to feed bytes into the chunker in reverse (since `BodyChunkIpld` pages have to be hashed back to front).

In the case of `BodyChunkIpld::load_all_bytes`, the solution is simpler: we should deprecate the method and instead prefer a method that yields pages of bytes as an async stream (e.g., `BodyChunkIpld::stream` or something to that effect).

PR #582, opened 2023-08-08
feat: add Content-Length headers when writing to a sphere.

jsantell:
First pass on this; we could also add length when `link`ing, either the size of the pointer or `0`


cdata:
Just as a general comment, I want to call attention to https://github.com/subconsciousnetwork/noosphere/issues/498 . There isn't a strict path dependency here, but I think the overall priority if #498 is higher, and that a more wholesome solution to computing `Content-Length` would arise from fixing it.

cdata (approved):
LGTM

PR #586, opened 2023-08-10
feat: Introduce a streaming way to encode `BodyChunkIpld`. Fixes #498

### LINKED ISSUE #498: Potentially unbounded memory overhead using `BodyChunkIpld`

jsantell:
* Introduce streaming `BodyChunkIpld::encode` and `BodyChunkIpld::decode` methods.
* Streaming mechanisms store data in memory until a threshold is hit, which then stores to disk.
* Mark non-streaming functions `BodyChunkIpld::load_all_bytes` and `BodyChunkIpld::store_bytes` as deprecated.
* Remove `BodyChunkDecoder` (now implemented as `BodyChunkIpld::decode`.

Also `bytes` has been moved to a workspace dep


rust/noosphere-core/src/data/body_chunk.rs
```diff
@@ -70,4 +76,315 @@ impl BodyChunkIpld {
...
+        byte_length: u64,
+    },
+    Disk {
+        temp_dir: tempfile::TempDir,
+        chunk_count: usize,
+        byte_length: u64,
+    },
+}
+
+impl ChunkStorageImpl {
```
cdata, line 239: Please avoid abbreviations e.g., `Impl`

jsantell: Open to suggestions on `ChunkStorage` (:woozy_face: ) as well!

rust/noosphere-core/src/data/body_chunk.rs
```diff
@@ -70,4 +76,326 @@ impl BodyChunkIpld {
...
+}
+
+enum BlobVecData {
+    Memory {
+        chunks: Vec<Vec<u8>>,
+        chunk_count: usize,
+        byte_length: u64,
+    },
+    #[cfg(not(target_arch = "wasm32"))]
+    Disk {
```
cdata, line 238: If we used a `Store`-based construct as our fundamental primitive for this, we could be generic to both web and native. On web, the scratch space would be a object store in IndexedDB, and on native it would be a Sled tree.

cdata: Just observing that encoding a large file will be no more viable on the web than it is on the desktop (and in fact we are probably _even more_ sensitive to our memory ceiling on the web).

cdata (changes requested):
At a high level this looks like the right approach. I think we should revise it to leverage a `impl Store` that is provided by <caller who wants to do the encoding> that is considered a persistence scratch space; that way the solution can address the problem of finite resources for any environment where we have _some_ implementation for storage persistence.

Cargo.toml
```diff
@@ -19,6 +19,7 @@ resolver = "2"
 [workspace.dependencies]
 anyhow = { version = "1" }
 axum = { version = "^0.6.18" }
+bytes = { version = "^1" }
```
jsantell, line 22: todo: remove `^`

rust/noosphere-core/src/data/body_chunk.rs
jsantell: ya SGTM!

rust/noosphere-storage/src/implementation/web.rs
```diff
@@ -75,7 +75,16 @@ impl Storage for WebStorage {
     }
 }
 
-#[derive(Clone)]
+#[async_trait(?Send)]
+impl Scratch for WebStorage {
+    type KeyValueStore = WebStore;
+
+    async fn get_scratch_store(&self) -> Result<Self::KeyValueStore> {
+        self.get_store("foo").await
```
jsantell, line 83: TODO: partition shared scratch space for indexedDB

rust/noosphere-storage/src/lib.rs
```diff
@@ -25,11 +26,12 @@ pub use encoding::*;
...
 pub use key_value::*;
 pub use retry::*;
+pub use scratch::*;
 pub use storage::*;
 pub use store::*;
 pub use tap::*;
 
-#[cfg(test)]
+#[cfg(feature = "helpers")]
 pub mod helpers;
```
jsantell, line 35: I don't think it's possible to export test-only exports between crates -- promoted to "helpers" here

rust/noosphere-storage/Cargo.toml
```diff
@@ -54,3 +55,7 @@ features = [
   "Window",
   "DedicatedWorkerGlobalScope",
 ]
+
+[features]
+default = ["helpers"]
+helpers = ["dep:witty-phrase-generator"]
```
jsantell, line 61: note: we could replace `witty-phrase-generator` with tempfile (native) and rand (web)

jsantell:
Updates:

* `NativeStorage`'s scratch space cleans up on drop
* `WebStorage`'s scratch space is partitioned under `SCRATCH_STORAGE` and wiped on storage instantiation. (If we could easily async-drop, other solutions are available)
* `Storage` now must implement `Scratch` -- we could roll these together (e.g. `Storage::get_scratch_space`), keep them independent (at the cost of `Storage + Scratch` everywhere), or keep this current middle ground.

rust/noosphere-storage/src/lib.rs
cdata: SGTM; I started doing this in `subconscious-cloud` the other day myself

rust/noosphere-core/src/data/body_chunk.rs
```diff
@@ -21,7 +27,7 @@ pub struct BodyChunkIpld {
 }
 
 impl BodyChunkIpld {
-    // TODO(#498): Re-write to address potentially unbounded memory overhead
+    #[deprecated]
```
cdata, line 30: Add deprecated note for both of these pointing to preferred alternative https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute

rust/noosphere-core/src/data/body_chunk.rs
```diff
@@ -70,4 +76,342 @@ impl BodyChunkIpld {
...
+            BODY_CHUNK_MAX_SIZE / 2,
+            BODY_CHUNK_MAX_SIZE,
+        );
+        let stream = chunker.as_stream();
+        tokio::pin!(stream);
+
+        let mut encoder = BodyChunkEncoder::new(store);
+        while let Some(chunk) = stream.try_next().await? {
+            encoder.push(chunk.data).await?;
+        }
```
cdata, line 98: It would be nice if the encoder wrapped the stream directly and only poll the stream if the encoder itself were being streamed (for maximum laziness).

Separately but related: it would be nice to be able to get the `BodyChunkIpld`s produced by the encoder as a stream. Streams are nice because you can express both "I want to do something at each step" and "I want to do something after all steps" with the same primitive.

rust/noosphere-core/src/data/body_chunk.rs
```diff
@@ -70,4 +76,342 @@ impl BodyChunkIpld {
...
+        let previous_stream = storage.into_stream(false).await;
+        tokio::pin!(previous_stream);
+        while let Some(chunk) = previous_stream.try_next().await? {
+            self.chunks.push(chunk).await?;
+        }
+        Ok(())
+    }
+}
+
+enum BlobVecData<S: KeyValueStore> {
```
cdata, line 247: I would say that the private implementation within this module deserves at least some top-level documentation for legibility's sake.

rust/noosphere-core/src/data/body_chunk.rs
```diff
@@ -70,4 +76,342 @@ impl BodyChunkIpld {
...
+    #[cfg(target_arch = "wasm32")]
+    wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_browser);
+
+    #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
+    #[cfg_attr(not(target_arch = "wasm32"), tokio::test)]
+    async fn it_reads_and_writes_chunks_nonstreaming() -> Result<()> {
+        let mut store = MemoryStore::default();
+        let chunk1 = vec![1; BODY_CHUNK_MAX_SIZE.try_into().unwrap()];
+        let chunk2 = vec![2; BODY_CHUNK_MAX_SIZE.try_into().unwrap()];
+        let chunk3 = vec![3; <u32 as TryInto::<usize>>::try_into(BODY_CHUNK_MAX_SIZE).unwrap() / 2];
```
cdata, line 378: It would be worth filling this test data with non-zero values, since a slice of zeros of the same size will get the same CID (and this could lend itself to a misleading outcome in the test).

rust/noosphere-storage/src/db.rs
```diff
@@ -33,9 +33,15 @@ pub const BLOCK_STORE: &str = "blocks";
 pub const LINK_STORE: &str = "links";
 pub const VERSION_STORE: &str = "versions";
 pub const METADATA_STORE: &str = "metadata";
+pub const SCRATCH_STORE: &str = "scratch";
 
-pub const SPHERE_DB_STORE_NAMES: &[&str] =
-    &[BLOCK_STORE, LINK_STORE, VERSION_STORE, METADATA_STORE];
+pub const SPHERE_DB_STORE_NAMES: &[&str] = &[
```
cdata, line 38: Strictly speaking I think this would call for a new version (and maybe migration, in `IndexedDB` terms). I don't think it's critical to do here, but something to think about. For example, anyone who used Sphere Viewer last Fall won't have a valid DB layout (for the given version) after this change.

rust/noosphere-storage/src/implementation/web.rs
```diff
@@ -179,3 +210,42 @@ impl Store for WebStore {
         Ok(old_value)
     }
 }
+
+#[derive(Clone)]
+pub struct TempWebStore {
```
cdata, line 215: Some docs on this would be nice

cdata: Ditto for the "native" equivalent.

rust/noosphere-core/src/data/body_chunk.rs
jsantell: The `vec![1; SIZE]` format uses the first value as an initializer, so these chunks are `[1, 1, 1, 1, ..]`, `[2, 2, 2, ..]` etc

rust/noosphere-core/src/data/body_chunk.rs
```diff
@@ -70,4 +76,342 @@ impl BodyChunkIpld {
...
+                *chunk_count += 1;
+            }
+            BlobVecData::Db {
+                store,
+                ref mut chunk_count,
+                ref mut byte_length,
+                ..
+            } => {
+                *byte_length += chunk.len() as u64;
+                let key = format!("{}", chunk_count);
```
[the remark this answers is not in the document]
jsantell: No guarantee, not dissimilar from any other `Store`. I'll add a note to `Scratch` that temporary stores specifically do not share state. A collision here would be no different than a `Storage`'s link store using data from its key value store, and implementation-wise, only IndexedDb's scratch store has a shared tree/table (where data bleed is tested in `SphereDb<S>`)

rust/noosphere-storage/src/db.rs
jsantell: Looks like Rexie does some upgrade handling for e.g. adding object stores. Incrementing to version 2, and adding a quick test

rust/noosphere-core/src/stream.rs
```diff
@@ -0,0 +1,216 @@
...
+    memory_limit: u64,
+) -> impl Stream<Item = Result<T, anyhow::Error>> + Unpin + 'a
+where
+    T: Reversable + 'static,
+    St: Stream<Item = Result<T, E>> + 'a,
+    S: Scratch,
+    E: Into<anyhow::Error>,
+{
+    Box::pin(try_stream! {
+        tokio::pin!(stream);
```
jsantell, line 23: TODO: is this needed?

rust/noosphere-core/src/stream.rs
```diff
@@ -0,0 +1,216 @@
...
+                    let mut db_store_inner = ReversableStore::new(provider.get_scratch_store().await?);
+                    db_store_inner.from_store(store).await?;
+                    db_store = Some(db_store_inner);
+                }
+            }
+        }
+
+        if let Some(store) = db_store.take() {
+            let output = store.into_reverse_stream().await;
+            tokio::pin!(output);
```
jsantell, line 44: TODO is this needed?

jsantell:
The tests were timing out the wasm bindgen test runner (20s per crate) due to large size; decreased the size of unit tests (with customizable memory limit) and started work on a benchmark for future improvements in the encoding time

rust/noosphere-storage/src/implementation/native.rs
```diff
@@ -99,3 +98,57 @@ impl Drop for NativeStorage {
...
+    }
+
+    async fn flush(&self) -> Result<()> {
+        self.store.flush().await
+    }
+}
+
+impl Drop for TempNativeStore {
+    fn drop(&mut self) {
+        if let Err(e) = self.db.drop_tree(&self.name) {
```
cdata, line 150: I wonder if we could better enforce this "storage disappears when dropped" aspect by saying something like:

```rust
pub trait DroppableStore {
  async fn drop_store(&mut self);
}

pub struct TemporaryStore(T) where T: Store + DroppableStore;

impl Store for TemporaryStore { /* ... */ }

impl DroppableStore for TemporaryStore {
  async fn drop_store(&mut self) {
    self.0.drop_store().await;
  }
}

impl Drop for TemporaryStore {
  fn drop(&mut self) {
    self.drop_store();
  }
}
```

And later we can implement async drop for `IndexedDbStore`. Then, we _also_ do a clean up check for droppable stores on every initialization of all storage layers, to cover the possibility that the application exited before async work could be scheduled.

cdata (commented):
Excited for this
giphy

rust/noosphere-core/examples/encoding.rs
```diff
@@ -0,0 +1,187 @@
...
+use noosphere_core::data::{BodyChunkIpld, BufferStrategy};
+use noosphere_core::tracing::initialize_tracing;
+use noosphere_storage::{helpers::make_disposable_storage, SphereDb, Storage};
+use std::collections::HashMap;
+use tokio::{self, io::AsyncRead};
+use tokio_stream::{Stream, StreamExt};
+use tokio_util::io::StreamReader;
+
+#[cfg(target_arch = "wasm32")]
+use web_time::Instant;
```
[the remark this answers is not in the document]
jsantell: `instant` is the way to go here :+1:

rust/noosphere-core/examples/encoding.rs
```diff
@@ -0,0 +1,187 @@
...
+
+#[derive(PartialEq, Debug)]
+enum BenchmarkPosition {
+    Start,
+    End,
+}
+
+/// Simple timer util to record duration of processing.
+/// Does not support nested, overlapping, or duplicate time ranges.
+struct EncodingBenchmark {
```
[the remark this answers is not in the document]
jsantell: Yes, similar constructs here; haven't updated this PR within that context, but some scaffolding can be shared here

rust/noosphere-storage/src/implementation/native.rs
jsantell: I think there's something here, but want to note that the `Drop` trait still needs a blocking executor

PR #590, opened 2023-08-10
chore: Clean up redundant dependencies

jsantell:
replace lingering async-std usage with tokio, promote tokio/tokio-stream to workspace dependency, remove unused deps, a few anyhows weren't using workspace etc.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ resolver = "2"
 
 [workspace.dependencies]
 anyhow = { version = "1" }
+async-stream = { version = "0.3" }
 axum = { version = "^0.6.18" }
 cid = { version = "0.10" }
 directories = { version = "5" }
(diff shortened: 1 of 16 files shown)


Cargo.toml
```diff
@@ -17,7 +17,8 @@ members = [
 resolver = "2"
 
 [workspace.dependencies]
-anyhow = { version = "1" }
+anyhow = { version = "^1" }
+async-stream = { version = "~0.3" }
```
cdata, line 21: I learned late that there are implicit semver ranges in Cargo that more or less mean what we intended here. So, I think we can remove `^` and `~`.

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio

cdata (dismissed):

cdata (approved):
Thanks for cleaning these up!

Issue #606, opened 2023-08-28
`LinkRecord` can propagate before blocks are available


cdata:
In https://github.com/subconsciousnetwork/noosphere/pull/253 we introduced a worker thread to help the gateway interact with the Noosphere Name System. This worker thread runs in parallel with the IPFS syndication worker thread introduced in #177.

This arrangement has worked well for small, incremental updates to spheres over time. However, in the case of a large update with many blocks, this arrangement results in an undesirable phase mismatch where the name system will receive an updated `LinkRecord` for a new sphere version before all of the blocks have managed to be syndicated to IPFS.

The end-user experience that results from this is that a user will try to traverse to a new version of a peer that should exist, but they won't be able to get the blocks for that version until some time in the future.

In my tests, the latency difference between the two processes is mainly bottlenecked by Kubo: one Kubo instance is only able to service syndication requests up to the number of cores on the local machine before it starts to queue requests and everything slows down (empirically / as observed from the outside, at any rate).

Perhaps the best solution to this problem is for name system and IPFS syndication tasks to work cooperatively / in serial. First, a syndication job is dispatched to the IFPS worker, and _then_ when that job is completed it forwards a name system publish job to the NNS worker. This would ensure that blocks are always available in IPFS before an updated name system record is published to peers. All other work to be done here is mainly tuning syndication and Kubo to reduce syndication latency as much as we can.

PR #608, opened 2023-08-28
fix: Increase allowed request body payload size

cdata:
This change increases our allowed body payload size to ~100MB. This is a stop-gap until such time as we update to streaming pushes (which can perhaps be treated in a special way). We also need to investigate:

- #606 
- #612
- #613

### DIFF
--- a/rust/noosphere-cli/tests/helpers/random.rs
+++ b/rust/noosphere-cli/tests/helpers/random.rs
@@ -1,0 +1,35 @@
+use rand::{rngs::StdRng, Rng, SeedableRng};
+use std::sync::Arc;
+use tokio::sync::Mutex;
+
+/// A helper to support consistent use of seeded randomness in tests.
+/// Its primary purpose is to retain and report the seed in use for a
+/// random number generator that can be shared across threads in tests.
+/// This is probably not suitable for use outside of tests.
+pub struct TestEntropy {
+    seed: [u8; 32],
+    rng: Arc<Mutex<StdRng>>,
+}
+
+impl Default for TestEntropy {
+    fn default() -> Self {
+        Self::from_seed(rand::thread_rng().gen::<[u8; 32]>())
+    }
+}
+
+impl TestEntropy {
+    pub fn from_seed(seed: [u8; 32]) -> Self {
+        tracing::info!(?seed, "Initializing test entropy...");
+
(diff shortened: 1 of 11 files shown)


rust/noosphere-cli/tests/peer_to_peer.rs
```diff
@@ -14,6 +14,7 @@ use noosphere_sphere::{
     SpherePetnameRead, SpherePetnameWrite, SphereReplicaRead, SphereSync, SphereWalker,
     SyncRecovery,
 };
+use rand::Rng;
```
justinabrahms, line 17: Any concerns w/ non-determinism in the test suite as a result of this?

justinabrahms (dismissed):

rust/noosphere-cli/tests/peer_to_peer.rs
cdata: Yes, although it's intentional in this case. The test writes to random keys (within a range) in the sphere. My intention is to simulate different probabilistic distributions.

We can enhance this by using a seed, and report the seed when running the test. At least that way we can reproduce the failure if we get a flake. I'll add that before we land this.

cdata: The RNG is now seeded, and the seed in use is reported in the test logs

rust/noosphere-cli/tests/helpers/random.rs
```diff
@@ -0,0 +1,35 @@
...
+
+impl Default for TestEntropy {
+    fn default() -> Self {
+        Self::from_seed(rand::thread_rng().gen::<[u8; 32]>())
+    }
+}
+
+impl TestEntropy {
+    pub fn from_seed(seed: [u8; 32]) -> Self {
+        tracing::info!(?seed, "Initializing test entropy...");
```
justinabrahms, line 22: I've not seen that ? Prefix before. Is it for optional params.

justinabrahms (approved):

rust/noosphere-cli/tests/helpers/random.rs
cdata: It's a `tracing`-specific syntax for logging just a value that implements `Debug`. It just appears as a field called `seed` in the log line.

Issue #607, opened 2023-08-28
Investigate SQLite- and RocksDB-based storage backends


cdata:
In very recent history we have run into a number of stability issues with Sled, our resident K/V storage backend that is used in all non-web-based clients today. Specifically: we have found ourselves stuck with a corrupted database that is unable to be opened, and at this time there is no recovery path.

It's possible that these stability issues are related to a known issue in Sled:

- https://github.com/spacejam/sled/issues/1351

Of course, we love using Sled for two important reasons:

1. It's a "pure Rust" dependency
2. It's very fast (possibly ~50% speed up compared to SQLite)
   - [Related commentary from the Sled author on performance trade-offs](https://github.com/spacejam/sled/issues/1444#issuecomment-1630321376) 

In the end, no amount of performance gain is worth the high-likelihood of database corruption, so we should begin exploring alternatives.

The two best candidates at this time are:

- [RocksDB](https://crates.io/crates/rusqlite), a Sled-like K/V store written in C++ ([Rust bindings](https://crates.io/crates/rocksdb))
- [SQLite](https://www.sqlite.org/) ([Rust bindings](https://crates.io/crates/rusqlite))

In the case of both of these alternatives, we will be taking on a major non-Rust dependency in the stack. And, in the case of SQLite in particular, we will be facing performance trade-offs (mainly to the downside) in exchange for apparent stability.

At any rate, let's implement storage for both of these backends along with subjective and objective tests to measure performance (and, if possible, reliability) of them compared to Sled. It is probably ideal for us to switch to RocksDB if it can be shown to be more stable, because it is almost certainly faster than SQLite. But, if the subjective performance of SQLite is indistinguishable from other backends, then the extra stability would give us some peace of mind.

### Aside

Honorable mention for potential alternatives goes to https://github.com/mozilla/rkv, which attempts to offer different backends in the same way that we do. They have a "safe mode" backend which perhaps betrays that Mozilla deals with some of the same challenges that we have faced with this work area overall.

cdata:
Related work from @jsantell : https://github.com/subconsciousnetwork/noosphere/pull/549

PR #622, opened 2023-09-05
chore: Genercize storage across gateway and rename "default" storages

jsantell:
chore: Rename NativeStorage -> SledStorage, WebStorage -> IndexedDbStorage. Make storage generic over gateway. In preparation for exploring pluggable backends in #607

### DIFF
--- a/rust/noosphere-gateway/src/authority.rs
+++ b/rust/noosphere-gateway/src/authority.rs
@@ -11,8 +11,8 @@ use axum::{
 use libipld_core::cid::Cid;
 use noosphere_core::authority::{SphereAbility, SphereReference, SPHERE_SEMANTICS};
 use noosphere_sphere::SphereContext;
-use noosphere_storage::NativeStorage;
 
+use noosphere_storage::Storage;
 use tokio::sync::Mutex;
 use ucan::{capability::CapabilityView, chain::ProofChain, store::UcanJwtStore};
 
@@ -23,12 +23,16 @@ use super::GatewayScope;
 /// represented by a UCAN. Any request handler can use a GatewayAuthority
 /// to test if a required capability is satisfied by the authorization
 /// presented by the maker of the request.
-pub struct GatewayAuthority {
+pub struct GatewayAuthority<S> {
     proof: ProofChain,
     scope: GatewayScope,
+    marker: std::marker::PhantomData<S>,
 }
 
-impl GatewayAuthority {
+impl<S> GatewayAuthority<S>
+where
+    S: Storage + 'static,
+{
     pub fn try_authorize(
         &self,
         capability: &CapabilityView<SphereReference, SphereAbility>,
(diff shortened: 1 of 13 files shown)


rust/noosphere/src/storage.rs
```diff
@@ -42,22 +42,22 @@ impl From<StorageLayout> for PathBuf {
...
 #[cfg(not(target_arch = "wasm32"))]
-use noosphere_storage::{NativeStorage, NativeStorageInit};
+use noosphere_storage::{SledStorage, SledStorageInit};
 
 #[cfg(not(target_arch = "wasm32"))]
 impl StorageLayout {
-    pub async fn to_storage(&self) -> Result<NativeStorage> {
-        NativeStorage::new(NativeStorageInit::Path(PathBuf::from(self)))
+    pub async fn to_storage(&self) -> Result<SledStorage> {
+        SledStorage::new(SledStorageInit::Path(PathBuf::from(self)))
```
jsantell, line 50: This (and `make_disposable_store`) are the last non-generic storage invocations AFAICT. As `PlatformStorage` could be a non-primitive (like `IpfsStorage` wrapper), some options:

* `Storage` adds a `new_from_path` method that works on all Storage impls (or `new_from_str`)
* `noosphere::platform` exposes something like `make_platform_storage(&str)`, though we want to distinguish `PlatformStorage` (which may be wrapped by a non-primitive store) from the underlying storage primitive, though currently this would involve a lot of duplicate defs if we want to swap around backends during testing, currently apple/aarch64 and non-native-apple multiplied by storage backends

rust/noosphere/src/platform.rs
```diff
@@ -19,13 +19,13 @@ mod inner {
     pub type PlatformKeyStorage = InsecureKeyStorage;
 
     #[cfg(not(feature = "ipfs-storage"))]
-    pub type PlatformStorage = NativeStorage;
+    pub type PlatformStorage = SledStorage;
```
justinabrahms, line 22: What is `platform` in this case and why is it different than whatever the alternative is?

jsantell: `PlatformStorage` here is the default storage used for a given platform for (mostly?) sphere storage, where its (given this patch) `IndexedDbStorage` (previously `WebStorage`) on wasm32 builds, and `SledStorage` (previously `NativeStorage`) elsewhere (though apple/aarch64 builds here are distinct, mostly paving the way for apple-specific key enclaves) -- this is where we'd compose different backends for different platforms e.g. maybe `SledStorage` for desktop builds but `RocksDbStorage` on iOS

jsantell: Though note that there's no functional changes in this patch, the sled backend is still used for non-web storage everywhere

justinabrahms:
Gotcha. It's the storage for the "current platform" where in that's set
with feature flags.

rust/noosphere-gateway/src/authority.rs
```diff
@@ -52,16 +56,17 @@ impl GatewayAuthority {
 }
 
 #[async_trait]
-impl<S> FromRequestParts<S> for GatewayAuthority
+impl<S, St> FromRequestParts<St> for GatewayAuthority<S>
 where
-    S: Send + Sync,
+    St: Send + Sync,
+    S: Storage + 'static,
```
justinabrahms, line 62: Just noting that there is no t (you named it `St`) in `Send + Sync` and there are three in `Storage + 'static`, which will probably confuse people w/ naming.

justinabrahms (approved):

rust/noosphere-gateway/src/authority.rs
jsantell: Nice catch! The `St` type reflects the underlying axum `State` as rationale for naming, but you make a great point, will update to something less ambiguous

cdata: I personally wouldn't balk at a non-abbreviated identifier for the generic type.

cdata (approved):
LGTM, this is great. I'm going to ask that we don't land before #624 but I will help with the rebase.

jsantell:
> LGTM, this is great. I'm going to ask that we don't land before #624 but I will help with the rebase.

It'll be easier rebasing #623 (which was built off of/evolved from this PR) and closing this PR most likely

jsantell:
Closing in lieu of superceding work in #623

PR #623, opened 2023-09-06
feat: Initial MVP implementation of RocksDbStorage

jsantell:
First pass on a RocksDB backend.

Built on #622 with a few more changes to make storage agnostic in the remaining parts (e.g. helpers and noosphere::platform). See the [patch on top of #622](https://github.com/subconsciousnetwork/noosphere/compare/generic-storage...rocksdb?expand=1).

## Feature Flags

There's an *additive* flag for using RocksDB in multithreaded mode (rational: target features using base feature, cleaner especially given a default assumption of "sled" in a world where we could also have more backends).

Flags for `noosphere` and `noosphere_cli` just propagate to `noosphere_storage`. A different name (`experimental-rocksdb` vs `rocksdb`) used at the top level to denote experimentality, and ensure Cargo feature flags propagate correctly when using a workspace rather than applying everywhere applicable. With more RocksDB experience, we can most likely just use one (single or multi threaded).

* noosphere:experimental-rocksdb
* noosphere:experimental-rocksdb-mt
* noosphere_cli:experimental-rocksdb
* noosphere_cli:experimental-rocksdb-mt
* noosphere_storage:rocksdb
* noosphere_storage:rocksdb-mt

There are other ways of managing these backends, no strong preference.

## Performance

Definitely room for improvement here, but currently the RocksDB implementation is half as slow as sled. No noticeable difference between single and multithreaded RocksDB implementations.

Running full test suite (with inconsistent doc tests)
* 55s-70s RocksDB
* 28s-40s Sled

Running `noosphere-cli` tests only
* 5.9s RocksDB
* 2.5s Sled


jsantell:
Added `PerformanceStorage<S: Storage>` for tracking perf/space, a `Space` trait (with `Space::get_space_usage`) implemented for storage providers.

The benchmarks are in `noosphere_storage/examples/bench`. This work is gated behind `performance` flag, though even with features defined in the example manifest, it still needs passed in the command line. Updated CI to add these flags, but there's gotta be a better solution.

jsantell:
Initial takeaways of running some high-level comparisons in these benches:


| DB            | Sm Read | Lg Read | Sm Write | Lg Write | Space Amp |
| ------------- | ------- | ------- | -------- | -------- | --------- |
| Sled          | 10us    | 30us    | 30us     | 140us    | 2         |
| Rocks         | 4us     | 10us    | 30us     | 100us    | 1.1       |
| IDB (Firefox) | 700us   | 1200us  | 900us    | 1300us   | 1.2       | 
| IDB (Chrome)  | 1200us  | 1200us  | 7700us   | 9700us   | 1.2       |

* Not calling RocksDB flush results in similar read/write performance between Rocks and Sled (with Rocks slightly faster). The WAL should mean we do not need to manually call flush.
* Firefox write performance is faster than Chrome by an order of magnitude (1-1.5ms vs 8-10ms)
* RocksDB and IndexedDb implementations have pretty good space amplification (~1.25x). Sled is closer to 2x.

jsantell:
Adding `--experimental-rocksdb` to apple builds to test the generated artifact

jsantell:
Merged in SqliteStorage work from #549 (stats updated in https://github.com/subconsciousnetwork/noosphere/pull/623#issuecomment-1712257299), fixed up glue to hopefully generate a usable iOS simulator artifact

rust/noosphere-core/src/bench.rs
```diff
@@ -0,0 +1,182 @@
...
+        BenchStats {
+            name,
+            median_duration_ms: total_duration / iterations as u64,
+            median_bytes: total_bytes / iterations as u64,
+            iterations,
+        }
+    }
+}
+
+fn pad_string(base: &str, column_len: usize, front_pad: bool) -> String {
```
cdata, line 70: I think this might be covered by Rust format strings: https://doc.rust-lang.org/std/fmt/#fillalignment

rust/noosphere-core/src/bench.rs
```diff
@@ -0,0 +1,182 @@
...
+        #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
+        async fn noosphere_benchmark_wasm32_main() {
+            $crate::tracing::initialize_tracing(None);
+            let benches = noosphere_benchmark_collect_benches();
+            $crate::bench::run_benches(benches).await.unwrap();
+        }
+
+        #[cfg(not(target_arch = "wasm32"))]
+        #[tokio::main(flavor = "multi_thread")]
+        pub async fn main() {
```
cdata, line 176: Why is it that non-Wasm is a `main` entrypoint while Wasm is a test?

rust/noosphere-storage/src/implementation/indexed_db.rs
```diff
@@ -132,50 +147,105 @@ impl WebStore {
...
 
-        WebStore::finish_transaction(tx).await?;
+        IndexedDbStore::finish_transaction(tx).await?;
 
         Ok(old_value)
     }
 }
+
+#[cfg(feature = "performance")]
+fn js_value_to_error(value: wasm_bindgen::prelude::JsValue) -> anyhow::Error {
```
cdata, line 199: Maybe some `impl From<JsValue> for OurError` is called for here..? Then `OurError` can have some means of being `anyhow::Error`-compatible.

rust/noosphere-storage/src/implementation/memory.rs
```diff
@@ -130,3 +130,18 @@ impl Store for MemoryStore {
...
+
+#[cfg(feature = "performance")]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+impl crate::Space for MemoryStorage {
+    async fn get_space_usage(&self) -> Result<u64> {
+        let mut size = 0;
+        for (_, store) in self.stores.lock().await.iter() {
+            for (_, entry) in store.entries.lock().await.iter() {
+                size += entry.len() as u64;
```
cdata, line 142: Technically the key is taking up space as well, no?

rust/noosphere-storage/src/implementation/performance.rs
```diff
@@ -0,0 +1,211 @@
...
+
+        let mut stats = PerformanceStats::default();
+        stats.reads = agg.reads;
+        stats.writes = agg.writes;
+        stats.removes = agg.removes;
+        stats.flushes = agg.flushes;
+        stats.median_reads_us = if agg.reads == 0 {
+            0
+        } else {
+            agg.reads_duration_us / agg.reads as u128
```
cdata, line 101: Is it acceptable to be doing integer division here? Don't you want floating point precision (and rounding as a formatting detail at the end)?

rust/noosphere-storage/src/implementation/rocks_db.rs
```diff
@@ -0,0 +1,158 @@
...
+#[cfg(feature = "rocksdb-mt")]
+type DbInner = DBWithThreadMode<rocksdb::MultiThreaded>;
+#[cfg(feature = "rocksdb-mt")]
+type ColumnType<'a> = Arc<rocksdb::BoundColumnFamily<'a>>;
+
+/// A RocksDB implementation of [Storage].
+///
+/// Caveats:
+/// * Values are limited to 4GB(?) [https://github.com/facebook/rocksdb/wiki/Basic-Operations#reads]
+/// * Potential workarounds of cf_handles not being [Sync] [https://github.com/rust-rocksdb/rust-rocksdb/issues/407]
```
cdata, line 20: `Unshared` in #624

rust/noosphere-storage/src/implementation/rocks_db.rs
```diff
@@ -0,0 +1,158 @@
...
+        let cf = &cf;
+        // @TODO atomicize this
+        let old_bytes = self.db.get_cf(cf, key)?;
+        self.db.delete_cf(cf, key)?;
+        Ok(old_bytes)
+    }
+
+    async fn flush(&self) -> Result<()> {
+        // With the use of WAL, we do not want to actively flush on every sync,
+        // and instead allow RocksDB to determine when to flush to OS.
```
cdata, line 135: :+1:

rust/noosphere-storage/src/implementation/rocks_db.rs
```diff
@@ -0,0 +1,158 @@
...
+        #[cfg(feature = "rocksdb-mt")]
+        let cf = &cf;
+        Ok(self.db.get_cf(cf, key)?)
+    }
+
+    async fn write(&mut self, key: &[u8], bytes: &[u8]) -> Result<Option<Vec<u8>>> {
+        let cf = self.cf_handle()?;
+        #[cfg(feature = "rocksdb-mt")]
+        let cf = &cf;
+        // @TODO atomicize this
```
cdata, line 117: nit: `// TODO(#issue):`

rust/noosphere-storage/src/implementation/rocks_db.rs
```diff
@@ -0,0 +1,158 @@
...
+        // let cf = &cf;
+        // self.db.flush_cf(cf)?;
+
+        Ok(())
+    }
+}
+
+impl Drop for RocksDbStorage {
+    fn drop(&mut self) {
+        let _ = self.db.flush();
```
cdata, line 148: Is this desired given the comment about WAL above?

rust/noosphere-storage/src/implementation/sqlite.rs
```diff
@@ -0,0 +1,354 @@
...
+    }
+
+    #[derive(Clone, Debug)]
+    enum SqliteResponse {
+        Value(Option<Vec<u8>>),
+        Empty,
+    }
+
+    #[derive(Debug)]
+    pub struct SqliteClient {
```
cdata, line 136: I think we should compartmentalize the SQLite work until we can work through the implementation together (I'm still not sure why we are managing a separate thread to access it, for example).

cdata (commented):
Awesome work on this.

rust/noosphere-core/src/bench.rs
jsantell: wasm_bindgen_test is the "host" executable; can run in browsers, though limited on headless due to timeouts

jsantell:
Updated from feedback (not yet accounting for rebase shuffling of some helpers)

rust/noosphere-storage/src/implementation/rocks_db.rs
jsantell: RocksDB only has atomicity for writes, removing this todo

rust/noosphere-core/src/bench.rs
subconscious-ops: I guess what I'm reaching for is: do we need asymmetry here, or can we structure the "native" branch as a test also (thereby using a combined entrypoint for both cases)?

rust/noosphere-storage/src/implementation/rocks_db.rs
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    async fn flush(&self) -> Result<()> {
+        // With the use of WAL, we do not want to actively flush on every sync,
+        // and instead allow RocksDB to determine when to flush to OS.
+
+        // let cf = self.cf_handle()?;
+        // #[cfg(feature = "rocksdb-mt")]
+        // let cf = &cf;
+        // self.db.flush_cf(cf)?;
```
cdata, line 177: Is there a TODO to go with this commented code?

rust/noosphere/build.rs
```diff
@@ -0,0 +1,25 @@
+#![recursion_limit = "256"]
+use cfg_aliases::cfg_aliases;
+
+fn main() {
+    cfg_aliases! {
+        // Platforms
+        wasm: { target_arch = "wasm32" },
+        native: { not(target_arch = "wasm32") },
```
cdata, line 8: I often put "native" in scare quotes when I write it because if you talk to someone who doesn't do web dev, often "native" is a meaningless term to them. Also, there is an argument to be made that "native" is relative and therefor JS/Wasm is "native" to web.

No action required, but just always find myself wondering if there is a better term to use here..

rust/noosphere/src/platform.rs
```diff
@@ -3,12 +3,11 @@
 ///! secure key management. This module lays out the concrete strategies we will
 ///! use on a per-platform basis.
 
-#[cfg(all(
-    any(target_arch = "aarch64", target_arch = "x86_64"),
-    target_vendor = "apple"
-))]
+#[cfg(all(feature = "rocksdb", feature = "sqlite",))]
+compile_error!("Cannot enable multiple storage backends.");
```
cdata, line 7: FWIW Rust suggested practice is to gracefully accept features that ought to be mutually exclusive: https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features

cdata: Also: if we do go down the suggested route here, we need to include all the back-ends which today includes `sled` and `indexed_db`.

You might get better mileage by using conditional selection in the `noosphere/build.rs` crate you added so that native backends have a well-defined order of precedence (e.g., if `all(rocks, sled)` then rocks wins, and subsequently if `all(rocks, sqlite)` the SQLite wins; `wasm` results in any native backend being disregarded etc.).

rust/noosphere/build.rs
```diff
@@ -0,0 +1,25 @@
+#![recursion_limit = "256"]
+use cfg_aliases::cfg_aliases;
```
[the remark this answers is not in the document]
jsantell: Oops, this was for debugging a missing comma in the macro :upside_down_face: great catch

rust/noosphere/src/platform.rs
jsantell: SG, with `cfg_aliases` this becomes easy to have ordinal usage e.g. `--features rocksdb --features sqlite` would prefer e.g. rocksdb

rust/noosphere-storage/src/implementation/rocks_db.rs
jsantell: Just an example of flushing, we can toss it

rust/noosphere/build.rs
jsantell: `not-wasm` :upside_down_face:

rust/noosphere-storage/src/implementation/rocks_db.rs
```diff
@@ -0,0 +1,218 @@
...
+            return Err(anyhow!("No such store named {}", name));
+        }
+
+        RocksDbStore::new(self.db.clone(), name.to_owned())
+    }
+
+    /// Opens a database at canonicalized `path`. Attempts to get an
+    /// already existing open reference to avoid a RocksDB limitation.
+    fn open_db(path: PathBuf) -> Result<Arc<DbInner>> {
+        let mut db_handles = DB_HANDLES
```
jsantell, line 64: @cdata would like some :eyes:  on this weak ref handling

cdata:
LMK when you are at a good point and I'll give the rebase a go now that #624 is landed.

rust/noosphere-storage/src/implementation/rocks_db.rs
```diff
@@ -0,0 +1,212 @@
...
+#[derive(Clone, Debug)]
+pub struct RocksDbStorage {
+    db: Arc<DbInner>,
+    #[allow(unused)]
+    path: PathBuf,
+}
+
+impl RocksDbStorage {
+    pub fn new(path: PathBuf) -> Result<Self> {
+        std::fs::create_dir_all(&path)?;
```
cdata, line 40: Might be safer to canonicalize first?

jsantell: Cannot canonicalize a path that does not exist

rust/noosphere-core/src/bench.rs
```diff
@@ -0,0 +1,182 @@
...
+
+pub struct BenchDesc {
+    pub name: Cow<'static, str>,
+    pub func: Box<dyn Fn(&mut Bencher) -> BoxFuture<()>>,
+    pub iterations: u32,
+    pub ignore: bool,
+}
+
+#[derive(Default, Debug)]
+pub struct Bencher {
```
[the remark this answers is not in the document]
jsantell: Bumping this out of this release for now

rust/noosphere-storage/src/implementation/rocks_db.rs
jsantell: This has some challenges migrating to `Unshared<ColumnType<'a>>`, tracking a lifetime through a `Store` implementor (or to its relative `DbInner` instance), updating docs to add a note of Unshared

.github/workflows/noosphere_apple_build.yaml
```diff
@@ -72,7 +72,7 @@ jobs:
         with:
           brew: protobuf cmake
       - name: 'Cargo Build'
-        run: cargo build --package noosphere --release --locked --target ${{ matrix.target }}
+        run: cargo build --package noosphere --release --locked --target ${{ matrix.target }} --features rocksdb
```
jsantell, line 75: This is the only thing that should be backed out before landing.

rust/noosphere-common/src/channel.rs
```diff
@@ -31,14 +39,17 @@ impl From<RecvError> for ChannelError {
     }
 }
 
-/// Represents a request to be processed in `MessageProcessor`,
-/// sent from the associated `MessageClient`.
+/// Represents a request to be processed in [MessageProcessor],
+/// sent from the associated [MessageClient].
```
cdata, line 43: A doc example might be a nice stand-in for tests, just to demonstrate intended usage.

cdata: Ah, nevermind, I see there is a test!

.github/workflows/noosphere_apple_build.yaml
jsantell: Just decoupling the implementation from the "on" switch -- we could make it a part of the build matrix, or configure it in `noosphere::platform`

rust/noosphere/src/ffi/noosphere.rs
```diff
@@ -94,6 +94,9 @@ impl Drop for NsNoosphere {
 /// You can also initialize the ns_noosphere_t with an optional third
 /// argument: a URL string that refers to a Noosphere Gateway API somewhere on
 /// the network that one or more local spheres may have access to.
+///
+/// Note that storages (`global_storage_path`, `sphere_storage_path`) can only
+/// be opened by a single `ns_noosphere` at a time.
```
cdata, line 99: :+1:

rust/noosphere-storage/src/implementation/rocks_db.rs
```diff
@@ -0,0 +1,148 @@
...
+type DbInner = DBWithThreadMode<rocksdb::MultiThreaded>;
+#[cfg(feature = "rocksdb-multi-thread")]
+type ColumnType<'a> = Arc<rocksdb::BoundColumnFamily<'a>>;
+
+/// A RocksDB implementation of [Storage].
+///
+/// Caveats:
+/// * Values are limited to 4GB(?) [https://github.com/facebook/rocksdb/wiki/Basic-Operations#reads]
+/// * Potential workarounds of cf_handles not being [Sync] [https://github.com/rust-rocksdb/rust-rocksdb/issues/407]
+///   Could potentially use `noosphere_common::Unshared`?
```
cdata, line 24: Can we file an issue and put a `TODO(#)`?

.github/workflows/noosphere_apple_build.yaml
cdata: IDK if it's possible to distinguish e.g., M2/iPad from M2/MacBook Air in this way. We may need to talk it over a little bit to figure out the safe path forward (or face coming up with a migration playbook for orb users).

rust/noosphere-storage/src/implementation/rocks_db.rs
```diff
@@ -0,0 +1,148 @@
...
+#[derive(Clone, Debug)]
+pub struct RocksDbStorage {
+    db: Arc<DbInner>,
+    #[allow(unused)]
+    path: PathBuf,
+}
+
+impl RocksDbStorage {
+    pub fn new(path: PathBuf) -> Result<Self> {
```
cdata, line 33: What happens if the storage is initialized multiple times over the same location? Idempotent/safe to do?

jsantell: Same restrictions as sled e.g. only one `new()` per `path` allowed at a time, e.g. this fails:

```rs
    async fn it_can_open_concurrent_dbs() -> Result<()> {
        let tempdir = tempfile::TempDir::new()?;
        let storage_path = tempdir.path().to_owned();

        let storage_1 = RocksDbStorage::new(storage_path.clone())?;
        let mut store_1 = storage_1.get_key_value_store("links").await?;
        let storage_2 = RocksDbStorage::new(storage_path.clone())?;
        let mut store_2 = storage_2.get_key_value_store("links").await?;

        let bytes_1 = vec![0, 1, 2, 3];
        let bytes_2 = vec![4, 5, 6, 7];
        store_1.write(b"from 1", &bytes_1).await?;
        store_2.write(b"from 2", &bytes_2).await?;
        assert_eq!(store_1.read(b"from 2").await?.unwrap(), bytes_2);
        assert_eq!(store_2.read(b"from 1").await?.unwrap(), bytes_1);
        Ok(())
    }
```

cdata (approved):
LGTM, excellent change!

PR #624, opened 2023-09-08
feat!: Replace `Bundle` with CAR streams in push

cdata:
This change removes all usage of `Bundle` within our network code, replacing it with [CARv1](https://ipld.io/specs/transport/car/carv1/)-formatted byte streams. These byte streams are produced by literal `Stream`s, which enable us to send large amounts of data back and forth without necessarily ballooning memory (though still subject to buffering by the `Stream` consumer).

### `reqwest` streaming complications

One complicating factor I encountered in this change is that our HTTP client of choice (`reqwest`) brings some caveats when considering streaming requests.

First, it's important to note: we prefer `reqwest` because it is adaptive to our Wasm targets, preferring a `hyper`-based implementation on native targets but substituting a `fetch`-based  API on web.

#### `Sync` streams

The first challenge I hit is that all streams used in `reqwest` bodies must meet a `Sync` bound, and most of our streams do not rigorously enforce this bound at this time (it seemed plausible to me that we might be able to meet it with some refactoring). Although in theory it would be interesting to share a stream across many threads, where each thread polls the stream so that chunks could be processed in parallel, in practice we do not have this use case today. For details on _why_ this bound is present in `reqwest`, please follow the links in this issue I filed against the project: https://github.com/seanmonstar/reqwest/issues/1969

To resolve this issue, I have adapted a suggestion from a [discussion in Rust internals](https://internals.rust-lang.org/t/what-shall-sync-mean-across-an-await/12020/2). We now have a helper called `Unshared` that can wrap `!Sync` things and add an unsafe `Sync` implementation to them. The wrapper only allows mutable borrows of the wrapped value, which presumably are exclusive, justifying the `Sync` bound. Additionally I added an `UnsharedStream` wrapper that implements `Stream` and gives us a `Stream + Sync` for any `Stream + Send` that we wrap it with (via `Unshared`), enabling us to satisfy the bounds in `reqwest` streaming bodies.

#### Wasm streams

I encountered another complication with `reqwest`/Wasm: although streaming request bodies via `fetch` are possible thanks to [`ReadableStream`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream), `reqwest`/Wasm does not implement support for them.

So, to work around this, I added a conditional compilation branch to our client for _push_ requests on `wasm32` targets. The upshot is that on web, we use a different request method when pushing that uses `fetch` via `gloo-net` with a `ReadableStream` body.

### Crate refactor

This change ended up leading to a major refactor of our crates. Implementing streaming push means adding streaming support to our API client. However, we have (on `main`) an existing dependency edge from `noosphere-sphere` (where streaming is implemented) to `noosphere-api` (where the client is implemented). So, adding streaming support to the API client would have created a dependency cycle between our crates (which is not allowed). Originally, both `noosphere-api` and `noosphere-sphere` were inlined to the `noosphere-core` crate, so I decided to migrate us back to that state with this change as it resolves the cyclical dependency. The `noosphere-sphere` and `noosphere-api` crates will remain in-tree until we have the chance to publish a final, `+deprecated` version for both of them.

Breaking changes include:
- Removes `noosphere-sphere` and `noosphere-api` crates, consolidating them with `noosphere-core`
- Deprecates gateway route `/api/v0alpha1/push` in favor of `/api/v0alpha2/push`
- Deprecates all usages of `Bundle`
- Removes `SphereDb::put_block_stream` (replaced with more generalized helper)
- All test helpers are re-organized, now all appearing within `helpers` modules inside of their domain-relevant crates, and require the `helpers` feature to be used across crates
- `test_kubo` feature renamed to `test-kubo`


rust/noosphere-core/src/context/has.rs
```diff
@@ -0,0 +1,223 @@
...
+};
+use tokio::sync::{Mutex, OwnedMutexGuard};
+
+use crate::context::SphereContextKey;
+
+use super::SphereContext;
+
+#[allow(missing_docs)]
+#[cfg(not(target_arch = "wasm32"))]
+pub trait HasConditionalSendSync: Send + Sync {}
```
jsantell, line 21: Can these use the `ConditionalSend` and related traits from `noosphere-common`?

rust/noosphere-core/src/context/metadata.rs
```diff
@@ -0,0 +1,36 @@
+//! These constants represent the metadata keys used when a [SphereContext] is
+//! is initialized. Since these represent somewhat free-form key/values in the
+//! storage layer, we are make a best effort to document them here.
```
jsantell, line 3: nit: `we make a best effort`

rust/noosphere-core/src/context/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+//! spheres. If you have storage and network primitives on your platform, you
+//! can initialize a [SphereContext] and use it to work with and synchronize
+//! spheres, as well as traverse the broader Noosphere data graph.
+//!
+//! In order to initialize a [SphereContext], you need a [Did] (like an ID) for
+//! a Sphere, a [Storage] primitive and an [Author] (which represents whoever or
+//! whatever is trying to access the Sphere inquestion).
+//!
+//! Once you have a [SphereContext], you can begin reading from, writing to and
+//! traversing the Noosphere content graph.
```
jsantell, line 11: Excellent docs

rust/noosphere-core/src/context/petname/mod.rs
```diff
@@ -0,0 +1,10 @@
+//! Sphere petnames are shorthand names that are associated with DIDs. A petname
+//! can be any string, and always refers to a DID. The DID, in turn, may be
```
jsantell, line 2: nit: Petnames MAY refer to `None` when first initialized

rust/noosphere-core/src/data/address.rs
```diff
@@ -15,12 +15,15 @@ use super::{Did, IdentitiesIpld, Jwt, Link, MemoIpld};
 #[cfg(docs)]
 use crate::data::SphereIpld;
 
+/// The name of the fact (as defined for a [Ucan]) that contains the link for a
+/// [LinkRecord]
```
jsantell, line 19: nit: In the future, we should consider terminating punctuation on all doc comments

rust/noosphere-core/src/data/headers/version.rs
```diff
@@ -1,8 +1,11 @@
 use anyhow::anyhow;
 use std::{convert::Infallible, fmt::Display, str::FromStr};
 
+/// The Noosphere protocol version
```
jsantell, line 4: How does this correlate with API version e.g. `v0alpha2`?

rust/noosphere-core/src/stream/block.rs
```diff
@@ -0,0 +1,42 @@
...
+use libipld_cbor::DagCborCodec;
+use libipld_core::raw::RawCodec;
+use noosphere_storage::BlockStore;
+use tokio_stream::{Stream, StreamExt};
+
+/// Helper to put blocks from a [Stream] into any implementor of [BlockStore]
+///
+/// Implementation note: this is a stand-alone helper because defining this
+/// async function on a trait (necessitating use of `#[async_trait]`) creates an
+/// unergonomic `Send` bound on returned [Future].
```
jsantell, line 12: I think I've circumvented this caveat for cross platform streams recently (in storage changes, or body chunk stream work) -- will evaluate during the rebase process to see

rust/noosphere-core/src/view/sphere.rs
```diff
@@ -31,6 +31,10 @@ use noosphere_storage::{base64_decode, block_serialize, BlockStore, SphereDb, St
 
 use super::{address::AddressBook, Authority, Delegations, Identities, Revocations, Timeslice};
 
+/// An arbitrarily long value for sphere authorizations that should outlive the
+/// keys they authorize
+// TODO: Recent UCAN versions allow setting a null expiry; we should use that
```
jsantell, line 36: :+1:

rust/noosphere-gateway/Cargo.toml
```diff
@@ -33,8 +36,8 @@ tower = { workspace = true }
 tower-http = { workspace = true, features = ["cors", "trace"] }
 async-trait = "~0.1"
 async-stream = { workspace = true }
-tracing = { workspace = true }
 wnfs-namefilter = { version = "0.1.21" }
+wnfs-common = "=0.1.23"
```
jsantell, line 40: Any reason for the version pinning?

rust/noosphere-gateway/src/gateway.rs
```diff
@@ -3,19 +3,20 @@ use axum::extract::DefaultBodyLimit;
...
-use noosphere_sphere::HasMutableSphereContext;
 use noosphere_storage::Storage;
 use std::net::TcpListener;
 use tower_http::cors::{Any, CorsLayer};
 use tower_http::trace::TraceLayer;
 use url::Url;
 
-use noosphere_api::route::Route as GatewayRoute;
+use noosphere_core::api::{v0alpha1, v0alpha2};
+// use noosphere_core::api::route::Route as GatewayRoute;
```
jsantell, line 16: nit comment

rust/noosphere-gateway/src/gateway.rs
```diff
@@ -85,17 +86,31 @@ where
...
+            &v0alpha1::Route::Push.to_string(),
+            #[allow(deprecated)]
+            put(handlers::v0alpha1::push_route::<C, S>),
+        )
+        .route(
+            &v0alpha2::Route::Push.to_string(),
+            put(handlers::v0alpha2::push_route::<C, S>),
+        )
+        .route(
+            &v0alpha1::Route::Fetch.to_string(),
```
jsantell, line 111: Mixing `v0alpha1` and `v0alpha2` here (outside of the deprecated routes maybe) feels like a smell; could `v0alpha2` re-export `v0alpha1` routes where there is no change?

rust/noosphere-gateway/src/handlers/v0alpha1/fetch.rs
```diff
@@ -4,14 +4,15 @@ use anyhow::Result;
 
 use axum::{body::StreamBody, extract::Query, http::StatusCode, Extension};
 use bytes::Bytes;
-use noosphere_api::data::FetchParameters;
 use noosphere_core::{
+    api::v0alpha1::FetchParameters,
```
jsantell, line 8: similar to the routes, I'd expect the parameters to consistently come from whatever the current API version is e.g. `api::v0alpha2::FetchParameters`

rust/noosphere-gateway/src/handlers/v0alpha2/mod.rs
```diff
@@ -0,0 +1,3 @@
+mod push;
+
+pub use push::*;
```
jsantell, line 3: Similar to the API in `noosphere-core`, I'd expect re-exports, and import `handlers::v0alpha2` within `noosphere-gateway`

rust/noosphere-ipfs/src/storage.rs
```diff
@@ -36,24 +37,24 @@ where
     }
 }
 
-#[cfg(not(target_arch = "wasm32"))]
-pub trait IpfsStorageConditionalSendSync: Send + Sync {}
+// #[cfg(not(target_arch = "wasm32"))]
+// pub trait IpfsStorageConditionalSendSync: Send + Sync {}
```
jsantell, line 41: nit: comments

rust/noosphere/Cargo.toml
```diff
@@ -39,15 +40,20 @@ libipld-cbor = { workspace = true }
...
 
 [dev-dependencies]
 libipld-core = { workspace = true }
+rand = { workspace = true }
+serde_json = { workspace = true }
+noosphere-cli = { version = "0.14.1", path = "../noosphere-cli", features = ["helpers"] }
+noosphere-core = { version = "0.15.1", path = "../noosphere-core", features = ["helpers"] }
+noosphere-common = { version = "0.1.0", path = "../noosphere-common", features = ["helpers"] }
+noosphere-gateway = { version = "0.8.1", path = "../noosphere-gateway" }
+noosphere-ns = { version = "0.10.1", path = "../noosphere-ns" }
```
jsantell, line 56: native only dependency?

rust/noosphere/tests/cli.rs
```diff
@@ -0,0 +1,155 @@
+#![cfg(all(feature = "test-kubo", not(target_arch = "wasm32")))]
+
+//! Integration tests to demonstrate that the Noosphere CLI, aka "orb", works
+//! end-to-end in concert with the name system and backing block syndication
```
jsantell, line 4: :tada:

rust/noosphere/tests/distributed_stress.rs
```diff
@@ -1,168 +1,146 @@
-#![cfg(not(target_arch = "wasm32"))]
+#![cfg(all(feature = "test-kubo", not(target_arch = "wasm32")))]
 
-mod helpers;
+//! Integration tests that expect "full stack" Noosphere to be available, including
+//! name system and block syndication backend (e.g., IPFS Kubo). The tests in this
+//! module represent sophisticated, complicated, nuanced or high-latency scenarios.
```
jsantell, line 5: This test cleanup is great

rust/noosphere/tests/gateway.rs
```diff
@@ -12,23 +16,21 @@ use tokio::io::AsyncReadExt;
 use tokio_stream::StreamExt;
 use url::Url;
 
-use noosphere_api::route::Route;
+use noosphere_core::api::v0alpha1;
```
jsantell, line 19: Expecting `v0alpha2` here, an internal detail to know which route has/not changed between versions

rust/noosphere/tests/gateway.rs
```diff
@@ -388,6 +390,60 @@ async fn gateway_receives_sphere_revisions_from_a_client() -> Result<()> {
     Ok(())
 }
 
+// #[tokio::test]
```
jsantell, line 393: missing a TODO, or could make as a pending test

jsantell (dismissed):
Great work! Mostly nit comments, main suggestions are to only use a single API export from `noosphere-core` (e.g. `v0alpha2`) throughout the codebase (via re-exporting `v0alpha1` params and routes inside v0alpha2`, or through other means)

rust/noosphere-core/src/context/authority/mod.rs
```diff
@@ -0,0 +1,5 @@
+mod read;
```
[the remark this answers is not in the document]
cdata: It doesn't. Mainly `noosphere/src/stream` is worth your attention (some of it was moved out of `noosphere-context`)

rust/noosphere-core/src/data/headers/version.rs
cdata: Mmm, good point. I guess we would typically want to bump the protocol version for this change. I'll have to think about that.

rust/noosphere-gateway/Cargo.toml
cdata: Yes, actually. I should have left a comment, but WNFS deprecated the `wnfs-namefilter` crate quietly and then broke compatibility in `wnfs-common`. Pinning to this version gives us breathing room to migrate.

rust/noosphere-gateway/src/gateway.rs
cdata: I guess it's a question of style, but the distinction is useful because we mount *both* `v0alpha1` and `v0alpha2` routes (see a few lines above) for push, which let's us enable clients to upgrade without breaking in-between.

rust/noosphere-gateway/src/handlers/v0alpha1/fetch.rs
cdata: See comment above.

rust/noosphere-gateway/src/handlers/v0alpha2/mod.rs
cdata: See comment above. The user-visible manifestation of these routes will be that there are routes at `v0alpha1/*` and `v0alpha2/*`, which is being done intentionally for migration/compatibility. It seems fitting that our code organization mirrors this ground truth.

rust/noosphere-common/src/sync.rs
```diff
@@ -0,0 +1,27 @@
+#[allow(missing_docs)]
+#[cfg(not(target_arch = "wasm32"))]
+pub trait ConditionalSend: Send {}
```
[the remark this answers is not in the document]
cdata: I like something like `Adaptive` or `Conditional`, because the word actually conveys semantics. `Platform` leaves it up to the imagination.

rust/noosphere-core/src/authority/author.rs
```diff
@@ -20,7 +20,9 @@ use super::generate_capability;
 /// else read-only access (to all other spheres).
 #[derive(PartialEq, Eq, Debug, Clone)]
 pub enum Access {
```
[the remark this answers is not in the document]
cdata: Well `SimulatedSphereAccess` is actually domain-specific to that helper. So, in tests where you use that helper, you need to use `SimulatedSphereAccess` (because it configures what the helper does when it produces a sphere for you).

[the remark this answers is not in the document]
jsantell: Context: generating a non-MemoryStore DB-backed sphere in tests https://github.com/subconsciousnetwork/noosphere/pull/623/files#diff-da1f415011687e703e572ced0f53a376dc3707ee9e6ccedf24eddea48c72a8a5R55

rust/noosphere-core/src/context/authority/mod.rs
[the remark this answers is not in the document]
jsantell: taking a look!

rust/noosphere-gateway/src/gateway.rs
jsantell: Backwards compat is for sure a larger topic to tackle, I need to think about this more

rust/noosphere/tests/gateway.rs
cdata: This test is included in this change, I just moved it somewhere else.

rust/noosphere-core/src/data/address.rs
cdata: Yah, I wonder if there is a spell check / style lint we can make use of for this kind of thing...

rust/noosphere-core/src/data/headers/version.rs
cdata: I'm going to say for now that this version concerns the structure of the data, and is at a different layer from the transport (e.g., Gateway REST API). But, let's discuss this more in detail. https://github.com/subconsciousnetwork/noosphere/issues/625

rust/noosphere-gateway/Cargo.toml
cdata: https://github.com/subconsciousnetwork/noosphere/issues/626

PR #633, opened 2023-09-13
feat: Introduce Importable/Exportable Storage traits.

jsantell:


rust/noosphere-storage/src/exportable.rs
```diff
@@ -0,0 +1,176 @@
...
+
+#[cfg(target_arch = "wasm32")]
+pub type IterableStream<'a> = Pin<Box<dyn Stream<Item = Result<(Vec<u8>, Option<Vec<u8>>)>> + 'a>>;
+#[cfg(not(target_arch = "wasm32"))]
+pub type IterableStream<'a> =
+    Pin<Box<dyn Stream<Item = Result<(Vec<u8>, Option<Vec<u8>>)>> + Send + 'a>>;
+
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+pub trait Iterable: Store {
```
jsantell, line 16: `Iterable` is an overloaded name -- while we could take the approach of `std`, where generic-named features are referenced via module (`io::Error`, `collections::hash_map::Iter`), maybe here we can append `Store` and `Storage` to most of our storage-specific traits e.g. `Space`.

TODO: Change to `IterableStore`

rust/noosphere-storage/src/exportable.rs
```diff
@@ -0,0 +1,176 @@
...
+}
+
+/// A blanket implementation for all [Storage]s to import
+/// an [Exportable] storage.
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+pub trait Importable<'a, E>
+where
+    Self: Storage,
+    Self::KeyValueStore: Store,
```
jsantell, line 50: All `Store`s implement `KeyValueStore`, maybe `Store` should implement `as_block_store()`/`as_key_store()` rather than `Storage` distinguish different `Store` wrappers which aren't used internally

rust/noosphere-storage/src/exportable.rs
```diff
@@ -0,0 +1,176 @@
...
+use async_trait::async_trait;
+use noosphere_common::ConditionalSync;
+use std::pin::Pin;
+use tokio_stream::{Stream, StreamExt};
+
+#[cfg(target_arch = "wasm32")]
+pub type IterableStream<'a> = Pin<Box<dyn Stream<Item = Result<(Vec<u8>, Option<Vec<u8>>)>> + 'a>>;
+#[cfg(not(target_arch = "wasm32"))]
+pub type IterableStream<'a> =
+    Pin<Box<dyn Stream<Item = Result<(Vec<u8>, Option<Vec<u8>>)>> + Send + 'a>>;
```
jsantell, line 12: values that are `None` could just be skipped as well

rust/noosphere-storage/src/exportable.rs
```diff
@@ -0,0 +1,177 @@
...
+            #[cfg(not(target_arch = "wasm32"))]
+            let obj = Self { temp_dir, path };
+
+            Ok(obj)
+        }
+    }
+
+    /// wasm32: MemoryStorage -> IndexedDbStorage
+    /// native: SledStorage -> MemoryStorage
+    /// native+rocks: SledStorage -> RocksDbStorage
```
jsantell, line 130: Can probably make `IndexedDbStorage` exportable with https://docs.rs/rexie/latest/rexie/struct.KeyRange.html

rust/noosphere-storage/src/exportable.rs
jsantell: This proposal breaks down with some storage layers e.g. `IpfsStorage` which rely on distinguishing block from KV stores

rust/noosphere-storage/src/meta.rs
```diff
@@ -0,0 +1,62 @@
+use crate::Storage;
```
jsantell, line 1: Unused

jsantell:
* `ImportStorage` and `ExportStorage` traits for storages to pipe entries to another. A Storage is exportable if `<S as Storage>::KeyValueStore: IterableStore`. Any storage with a writable store `<S as Storage>::KeyValueStore: Store` can import. `MemoryStorage`, `IndexedDbStorage`, and `SledStorage` can be exported.
* A few generic storage operators: `OpenStorage`, `RenameStorage`, `RemoveStorage`. A generic open utility is valuable in tests alone, and other functions generic over storage, the others only used for migrating. A trait `FsBackedStorage` can be implemented, providing a blanket implementation to these operators.
* While `IndexedDbStorage` is exportable, a storage rename (hard copy) isn't ideal. While there are tests for exporting indexedDB, not yet handling migrating on the web without a destination format in mind (other than `MemoryStore`)
* The remaining bits are probing existing storage to learn the type/version if it exists to explicitly use DB types during migration.
* Names are murky for now; open to suggestions on `FsBackedStorage`, and any consistencies of storage-based traits; e.g. `OpenStorage` etc. (verb+Storage) seems to fit the bill in most cases (whereas `Importable` and `Exportable` kind of mean the same thing)

jsantell:
Ready for review; would like to do the final bits in a follow up within `noosphere` where we determine if a migration is needed, what storage types are being involved, and any record keeping (db version/type), and actually enable the path

rust/noosphere-storage/src/migration.rs
```diff
@@ -0,0 +1,234 @@
...
+    <Self as Storage>::KeyValueStore: Store,
+{
+}
+
+impl<'a, S, T> MigratableTarget<'a, S> for T
+where
+    S: MigratableSource + ConditionalSync + 'a,
+    <S as Storage>::KeyValueStore: IterableStore,
+    Self: ImportStorage<'a, S> + Storage + OpenStorage + RemoveStorage + RenameStorage,
+    <Self as Storage>::KeyValueStore: Store,
```
jsantell, line 112: Unless `<Self as Storage>::KeyValueStore: Store` and ` <S as Storage>::KeyValueStore: IterableStore` can be inherited by a supertrait, `MigratableTarget` and `MigratableSource` can probably be folded away

rust/noosphere-storage/src/migration.rs
```diff
@@ -0,0 +1,234 @@
...
+    <S as Storage>::KeyValueStore: IterableStore,
+{
+    let temp_dir = tempfile::TempDir::new()?;
+    let temp_path = temp_dir.path();
+    {
+        let mut to_storage = T::open(temp_path).await?;
+        let from_storage = S::open(path.as_ref()).await?;
+        to_storage.import(&from_storage).await?;
+    }
+    S::remove(path.as_ref()).await?;
```
jsantell, line 134: Create a backup here

PR #635, opened 2023-09-13
chore: Move benchmark to workspace-level example

jsantell:
This approach has the benefit of being an isolated crate and not inheriting from its host crate, and being more of an independent executable. It is checked/compiled at the workspace level (enabling performance-gated functionality linting on CI) while properly applying its feature flag needed (removing the need for performance feature enabled everywhere).

Supercedes the workarounds needed in #634


cdata:
All things considered, I think I would prefer this be an example in an existing gated by feature flags (even if it is ugly to express that). Our crate space is busy enough as it is, and we have prior experience doing this kind of thing (e.g., for `noosphere-cli` where we go to pains to make it spit out nothing when targeting web).

An alternative I might consider is making a new crate that is specifically themed after performance / benchmarking, where all such code can live.

jsantell:
> All things considered, I think I would prefer this be an example in an existing gated by feature flags (even if it is ugly to express that). Our crate space is busy enough as it is, and we have prior experience doing this kind of thing (e.g., for `noosphere-cli` where we go to pains to make it spit out nothing when targeting web).

Desired qualities of a standalone executable here, with consequences being more annoying than disastrous:
1) minimal macro boilerplate
2) compile-checked/linted when testing/building the workspace even if not executed
3) does not pollute workspace with its feature flags
4) "portable": crate example crates inherit host deps/configs  -- not immediately valuable for this case, but a crate that can be dropped (swap out workspace deps) into a new dir and `cargo run` is valuable for examples/3P

The approach in https://github.com/subconsciousnetwork/noosphere/pull/634/files describes your suggestion, only having the qualities of (3).

Keeping a limit on published `noosphere-*` crates is something to keep an eye on for sure, though `rust/examples/*` breaks the mold a bit without being a published crate (nor containing publishable crates), using a well-known cargo dir. Could be placed at the root as well to minimize `rust/*` directories. Some advantage to having a general-purpose `examples` dir as a catch-all for various use cases, especially when they don't quite line up with one single `noosphere-` crate (e.g. transitive workflows).

> An alternative I might consider is making a new crate that is specifically themed after performance / benchmarking, where all such code can live.

This solution achieves (1), (2), and (3), at the cost of a new published crate (e.g. noosphere-performance). If we want to limit published noosphere crates, this is pushing against that and might not work for the next solution that doesn't match "performance".

---

As `StoragePerformance` is currently only used by the benchmark tests, we can move that alongside the bench example (in noosphere-storage) for now (again achieving (1), (2), and (3)), while avoiding new dirs in `rust/*` and any published crates, side-stepping this unless we need multiple uses of feature flagged behavior

jsantell:
> As `StoragePerformance` is currently only used by the benchmark tests, we can move that alongside the bench example (in noosphere-storage) for now (again achieving (1), (2), and (3)), while avoiding new dirs in `rust/*` and any published crates, side-stepping this unless we need multiple uses of feature flagged behavior

Trying this approach in https://github.com/subconsciousnetwork/noosphere/pull/636

PR #636, opened 2023-09-14
chore: Move PerformanceStorage to benchmark example, remove need for performance feature flag

jsantell:
Other small changes:
* Deserializing in wasm-bindgen causes a lint warning to use `serde-wasm-bindgen`, applied here
* Use sled's API for getting size on disk. A few bytes short of the crude fs crawling.
* `noosphere_storage::Space` is no longer feature gated.
* `instant` able to be moved to storage's dev-dependencies -- the bench example `Cargo.toml` isn't even being used! Not sure if it can support freeform dependencies like non-example-hosted crates like #635


.github/workflows/run_test_suite.yaml
```diff
@@ -55,7 +55,7 @@ jobs:
           ipfs_version: v0.17.0
           run_daemon: true
       - name: 'Run Rust native target tests'
-        run: cargo test --features test-kubo,helpers,performance
+        run: cargo test --features test-kubo,helpers
```
cdata, line 58: It might be interesting to run the benchmark in its own job, just to make sure we're keeping it in a working state.

jsantell: that's not necessary, it is still compiled and linted in CI

cdata (approved):
LGTM

PR #643, opened 2023-09-19
feat: Do not proactively create directories when creating spheres via CLI. Fixes #599

### LINKED ISSUE #599: validate key exists before creating `.sphere`
If you attempt to create a sphere w/ an invalid key, it creates the `.sphere` directory which prevents subsequent attempts unless you remove the directory.

```
[abrahms@subconscious-nix:/tmp/tmp.q1VCtOS7or]$ ls -alh
total 7.5M
drwx------   2 abrahms users 4.0K Aug 17 11:32 .
drwxrwxrwt 173 root    root  7.4M Aug 17 11:32 ..

[abrahms@subconscious-nix:/tmp/tmp.q1VCtOS7or]$ orb sphere create --owner-key jimmy-john
Error: No key named jimmy-john found!

[abrahms@subconscious-nix:/tmp/tmp.q1VCtOS7or]$ ls -alh
total 7.5M
drwx------   3 abrahms users 4.0K Aug 17 11:33 .
drwxrwxrwt 173 root    root  7.4M Aug 17 11:32 ..
drwxr-xr-x   6 abrahms users 4.0K Aug 17 11:33 .sphere

[abrahms@subconscious-nix:/tmp/tmp.q1VCtOS7or]$ orb sphere create --owner-key user1
Error: A sphere is already initialized in /tmp/tmp.q1VCtOS7or
```

jsantell:


cdata:
Can you provide a rationale for the change? Maybe a test?

rust/noosphere/src/sphere/builder.rs
```diff
@@ -166,6 +166,10 @@ impl SphereContextBuilder {
         let storage_path = self
             .storage_path
             .ok_or_else(|| anyhow!("No storage path configured!"))?;
+
+        #[cfg(not(target_arch = "wasm32"))]
+        std::fs::create_dir_all(&storage_path)?;
```
cdata, line 171: I think we're reaching the long way around this problem here. Instead of avoiding creating a folder, would it be possible to check that a the key exists at the appropriate time?

jsantell:
> Can you provide a rationale for the change? Maybe a test?

In #599, for early failures (like an invalid key), the SpherePaths initialization proactively creates all of the `.sphere` dirs., preventing follow up commands with the correct key, as the sphere dirs are no longer uninitialized. Added a test in `noosphere/tests/cli.rs`

cdata:
Thanks, I missed that the bug was in the title at first :+1:

rust/noosphere/tests/cli.rs
```diff
@@ -37,6 +37,28 @@ async fn orb_sphere_create_initializes_a_sphere() -> Result<()> {
     Ok(())
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn orb_sphere_create_failure_cleans_directory() -> Result<()> {
```
cdata, line 41: :100:

rust/noosphere/src/sphere/builder.rs
jsantell: This was over-aggressive on my end, and not needed for any of our FS-backed Storages

PR #646, opened 2023-09-20
chore: Add some RocksDB-related tests

cdata:


cdata:
Going to close this out for now. We can always revisit.

PR #661, opened 2023-10-03
feat: History compaction for spheres and gateways

cdata:
This change proposes a form of incremental history compaction, the core of which is implemented in `Sphere`. Its only use at this time will be in the Noosphere Gateway, where it will be used to ensure that side-effects that cause history to accrue for the gateway's sphere do not result in unbounded version growth. As of this change, the gateway will attempt to compact its history once an hour. The compaction flattens all history since the last user sync point.

### DIFF
--- a/rust/noosphere-core/src/view/mutation.rs
+++ b/rust/noosphere-core/src/view/mutation.rs
@@ -194,32 +194,14 @@ impl SphereMutation {
 
     /// Consume a [SphereMutation], appending its changes to this one
     pub fn append(&mut self, other: SphereMutation) {
-        self.content.changes = append_changes(
-            std::mem::take(&mut self.content.changes),
-            other.content.changes,
-        );
-
-        self.identities.changes = append_changes(
-            std::mem::take(&mut self.identities.changes),
-            other.identities.changes,
-        );
-
-        self.delegations.changes = append_changes(
-            std::mem::take(&mut self.delegations.changes),
-            other.delegations.changes,
-        );
-
-        self.revocations.changes = append_changes(
-            std::mem::take(&mut self.revocations.changes),
-            other.revocations.changes,
-        );
+        append_changes(&mut self.content.changes, other.content.changes);
+        append_changes(&mut self.identities.changes, other.identities.changes);
+        append_changes(&mut self.delegations.changes, other.delegations.changes);
+        append_changes(&mut self.revocations.changes, other.revocations.changes);
     }
 }
 
-fn append_changes<K, V>(
-    mut destination: Vec<MapOperation<K, V>>,
-    source: Vec<MapOperation<K, V>>,
-) -> Vec<MapOperation<K, V>>
+fn append_changes<K, V>(destination: &mut Vec<MapOperation<K, V>>, source: Vec<MapOperation<K, V>>)
 where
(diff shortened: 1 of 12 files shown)


rust/noosphere-gateway/src/worker/name_system.rs
```diff
@@ -232,8 +232,13 @@ where
                 context,
                 republish,
             } => {
-                if let Err(error) = set_counterpart_record(context, &record).await {
-                    warn!("Could not set counterpart record on sphere: {error}");
+                // NOTE: Very important not to update this record on every
+                // re-publish, otherwise we will generate a new sphere revision
+                // on an on-going basis indefinitely
+                if !republish {
```
jsantell, line 238: :exploding_head: great catch. the test here ensures we handle republishing w/r/t expiry, but does not consider the effects on setting counterpart. Might be able to easily extend the test and check counterpart record

rust/noosphere-core/src/view/mutation.rs
```diff
@@ -191,6 +191,58 @@ impl SphereMutation {
             && self.delegations.changes.len() == 0
             && self.revocations.changes.len() == 0
     }
+
+    /// Consume a [SphereMutation], appending its changes to this one
+    pub fn append(&mut self, other: SphereMutation) {
+        self.content.changes = append_changes(
+            std::mem::take(&mut self.content.changes),
```
jsantell, line 198: The destination arg and return value are the same object, do we need to `take` and reassign `self`'s changes?

```rs
append_changes(&mut self.content.changes, other.content.changes);
```

rust/noosphere-core/src/view/sphere.rs
```diff
@@ -1499,44 +1542,43 @@ mod tests {
...
+        (change_key, change_memo): (&str, &MemoIpld),
+    ) -> anyhow::Result<Link<MemoIpld>> {
+        let mut mutation = SphereMutation::new(author_did);
+        mutation.content_mut().set(
+            &change_key.into(),
+            &store
+                .save::<DagCborCodec, _>(change_memo)
+                .await
+                .unwrap()
+                .into(),
```
jsantell, line 1567: nit: `?` instead of `unwrap()`

rust/noosphere-gateway/src/worker/cleanup.rs
```diff
@@ -0,0 +1,297 @@
...
+use tokio::{
+    sync::mpsc::{unbounded_channel, UnboundedReceiver, UnboundedSender},
+    task::JoinHandle,
+};
+use tokio_stream::StreamExt;
+
+// Once an hour
+const PERIODIC_CLEANUP_INTERVAL_SECONDS: u64 = 60 * 60;
+
+#[derive(EnumDisplay)]
```
jsantell, line 19: TIL

jsantell (dismissed):
fantastic work! some comments & questions

rust/noosphere-gateway/src/worker/cleanup.rs
cdata: This is just an aliased `strum` thing; the exported name is `Display` and it conflicts with `std::fmt::Display` so I usually rename it.

PR #671, opened 2023-10-06
feat: Introduce "--storage-memory-cache-limit"

jsantell:
* Introduce `ConfigurableStorage` trait to open storage with generalized configurations (only `memory_cache_limit` at the moment)
* Introduce `--storage-memory-cache-limit` and `NoosphereStorage { config: NoosphereStorageConfig, path: NoosphereStoragePath }` for propagating these values from orb and rust API respectively into the sphere builder, which makes its way into the storage provider.
* Make `invoke_cli()` take a mutable Workspace (for setting storage config)


rust/noosphere/tests/cli.rs
```diff
@@ -66,7 +69,11 @@ async fn orb_can_enable_multiple_replicas_to_synchronize() -> Result<()> {
         .orb(&["sphere", "config", "set", "counterpart", &client_sphere_id])
         .await?;
 
-    let gateway_task = tokio::task::spawn(async move { gateway.orb(&["serve"]).await });
+    let gateway_task = tokio::task::spawn(async move {
+        gateway
+            .orb(&["serve --storage-memory-cache-limit 99999"])
```
jsantell, line 74: TODO Remove all changes here, did not test yet

rust/noosphere/src/storage.rs
```diff
@@ -82,3 +62,36 @@ fn get_scoped_path(path: &Path, scope: &Did) -> PathBuf {
...
+    let storage = {
+        use noosphere_storage::ConfigurableStorage;
+        PrimitiveStorage::open_with_config(&path, storage_config.unwrap_or_default()).await?
+    };
+
+    #[cfg(wasm)]
+    let storage = IndexedDbStorage::new(&path.to_string()).await?;
+
+    #[cfg(ipfs_storage)]
+    let storage = {
```
jsantell, line 86: This now enables a native build with `ipfs-storage` enabled, and bails if an `ipfs_gateway_url` has not been set as `PlatformStorage`

jsantell:
* Sled: Limits the system page [cache capacity](https://docs.rs/sled/latest/sled/struct.Config.html#method.cache_capacity).
* RocksDB: [Memory usage in RocksDB](https://github.com/facebook/rocksdb/wiki/Memory-usage-in-RocksDB) can be attributed to the block cache, indexes/filter blocks, and memtable. This config field limits the overall memtable cache usage across all column families (stores). The [block cache](https://github.com/facebook/rocksdb/wiki/Block-Cache) is 32MB by default. May need to tweak these settings more in the future.
* IndexedDb: Does not use this value. I imagine IndexedDb will eventually have some storage configurations where it'd make sense to implement `ConfigurationStorage` for it, ignoring the unsupported values.

Other ideas:
* A `supports_config()` method in the trait if we need to concretely know how supported a config is (`enum ConfigSupport { Unsupported, PartialSupport, Supported }`)
* Dumping internal stats from a Storage (RocksDB has a lot of insight here) via maybe a timer or gateway API

rust/noosphere-cli/src/native/cli.rs
```diff
@@ -61,6 +61,11 @@ pub enum OrbCommand {
         /// The port that the gateway should listen on
         #[clap(short, long, default_value = "4433")]
         port: u16,
+
+        /// If set, the amount of memory that the storage provider may use
+        /// for caching in bytes.
+        #[clap(long)]
```
cdata, line 67: Can we get a short arg for this? e.g., `short = 'S'` or something

rust/noosphere-cli/src/native/commands/serve.rs
```diff
@@ -1,25 +1,28 @@
...
     ipfs_api: Url,
     name_resolver_api: Url,
     cors_origin: Option<Url>,
-    workspace: &Workspace,
+    storage_config: Option<StorageConfig>,
+    workspace: &mut Workspace,
 ) -> Result<()> {
+    if let Some(storage_config) = storage_config {
+        workspace.release_sphere_context();
+        workspace.with_storage_config(storage_config);
```
cdata, line 23: It seems like we pass this arg all the way in here just to invoke the configuration right away. Perhaps we are doing this configuration in the wrong place. Why shouldn't the CLI handle this configuration? Why wouldn't it be provided at workspace instantiation time?

rust/noosphere-cli/src/native/commands/sphere/mod.rs
```diff
@@ -42,7 +42,7 @@ use url::Url;
 pub async fn sphere_create(owner_key: &str, workspace: &mut Workspace) -> Result<(Did, Mnemonic)> {
     workspace.ensure_sphere_uninitialized()?;
 
-    let sphere_paths = SpherePaths::intialize(workspace.working_directory()).await?;
+    let sphere_paths = SpherePaths::initialize(workspace.working_directory()).await?;
```
cdata, line 45: Nice catch

rust/noosphere-cli/src/native/helpers/cli.rs
```diff
@@ -150,7 +150,8 @@ impl CliSimulator {
         Ok(())
     }
 
-    fn parse_orb_command(&self, command: &[&str]) -> Result<Cli> {
+    /// Parse an orb command slice into a [Cli].
+    pub fn parse_orb_command(command: &[&str]) -> Result<Cli> {
```
cdata, line 154: Any particular reason why this was made public?

rust/noosphere-cli/src/native/workspace.rs
```diff
@@ -164,6 +167,13 @@ impl Workspace {
         Ok(())
     }
 
+    /// Use the provided `storage_config` when opening the [SphereContext].
+    /// Use [Workspace::release_sphere_context] after setting storage config
+    /// to ensure sphere is opened with these settings.
+    pub fn with_storage_config(&mut self, storage_config: StorageConfig) {
```
cdata, line 173: What is the use case for configuring this lazily e.g., after the `Workspace` has been constructed?

rust/noosphere-storage/src/implementation/sled.rs
```diff
@@ -54,6 +58,25 @@ impl Storage for SledStorage {
     }
 }
 
+#[async_trait]
+impl ConfigurableStorage for SledStorage {
```
cdata, line 62: It seems like this implementation is the same for both `SledStorage` and `RocksDbStorage`. Can it just be a general default implementation?

rust/noosphere/src/storage.rs
```diff
@@ -82,3 +62,36 @@ fn get_scoped_path(path: &Path, scope: &Did) -> PathBuf {
...
+    #[cfg(wasm)]
+    let storage = noosphere_storage::IndexedDbStorage::new(&layout.to_string()).await?;
+
+    #[cfg(ipfs_storage)]
+    let storage = {
+        let maybe_client = ipfs_gateway_url.map(|url| GatewayClient::new(url));
+        IpfsStorage::new(storage, maybe_client)
+    };
+
+    debug!("Creating platform storage: {:#?}", storage);
```
cdata, line 94: This log line might be better written before the storage is created (otherwise change to past tense e.g., "created" not "creating"; but, I think moving it is better so that you get appropriate context for any subsequent error).

cdata (commented):
Great work, thanks for adding this capability.

rust/noosphere-cli/src/native/workspace.rs
jsantell: due to `invoke_cli(Cli, Workspace)` (or the change in this PR, `invoke_cli(Cli, &mut Workspace)`) -- though we could have `invoke_cli(Cli)` and `invoke_cli_with_workspace(Cli, &mut Workspace)` for scenarios where the workspace needs to be provided, and then this becomes a construction-time config

rust/noosphere-storage/src/implementation/sled.rs
jsantell: I don't think so, they both just coincidentally have the core of their instantiation and setup in `Self::with_config` rather than calling `<Self as ConfigurableStorage>::open_with_config` from `Self::new`

rust/noosphere-cli/src/native/helpers/cli.rs
jsantell: As a part of tests for this, wanted to parse an orb command and get the storage to assert that some config was set, but had issues with the serialization barrier and also the config only being on the long-running-command `serve` -- I'll roll this back

rust/noosphere-cli/src/native/commands/serve.rs
jsantell: agreed, this should be handled in `invoke_cli`

rust/noosphere/src/storage.rs
jsantell: Changed to `created`, as we use `storage`'s Debug implementation to log storage type and extra configurations (could add more debugging to type if needed e.g. `trait StorageInfo { fn name() -> String }`

rust/noosphere-cli/src/native/cli.rs
jsantell: Punting on this for now, as this is a low-level operator command

rust/noosphere-cli/src/native/mod.rs
```diff
@@ -29,102 +25,135 @@ use self::{
...
-/// depending on those inputs).
-pub async fn invoke_cli(cli: Cli, mut workspace: Workspace) -> Result<()> {
+/// Invoke the CLI implementation imperatively.
+///
+/// This is the entrypoint used by orb when handling a command line invocation.
+/// The [Cli] is produced by parsing the command line arguments, and internally
+/// creates a new [Workspace] from the current working directory.
+///
+/// Use [invoke_cli_with_workspace] if using your own [Workspace].
+pub async fn invoke_cli<'a>(cli: Cli, context: &CliContext<'a>) -> Result<()> {
```
jsantell, line 60: `invoke_cli` now creates a `Workspace` from its args rather than accept one so that we can always have the correct storage configuration on instantiation (this is the bulk of the follow up changes)

cdata (approved):
Looks great, thanks for implementing this!

PR #673, opened 2023-10-10
feat: 3P replication fall-back and resilience

cdata:
This change proposes the following:

- A "fall back to local data" strategy for peers with new versions that fail to replicate
- Per-block latency guards on CAR streams to detect stalled Kubo queries
- Secondary Kubo pinning for orphaned blocks in Noosphere CAR streams
- Timeouts when looking blocks up from Kubo have been reduced (net ~16s to net ~10s)
- Improvements to `car` example utility to report orphaned or missing blocks in a CARv1

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1928,9 +1928,9 @@ dependencies = [
 
 [[package]]
 name = "gloo-timers"
-version = "0.2.6"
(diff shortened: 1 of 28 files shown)


jsantell (approved):
very nice! :tada:

PR #685, opened 2023-10-18
feat: Periodic syndication checks to IPFS Kubo

cdata:
- Adds an automatic attempt by the gateway to syndicate to Kubo at startup and on a 5 minute interval
- Moves CI tests to `cargo-nextest` with `--retries 5` for relatively lenient flake allowance
- Makes logging in `orb` more verbose
- Rust test suite coverage expanded to include MacOS 13

We now have a Github Workflow that reports test flakes in test runs. The report looks like this: 
image

### DIFF
--- a/.github/workflows/run_test_suite.yaml
+++ b/.github/workflows/run_test_suite.yaml
@@ -56,86 +56,73 @@ jobs:
       - name: 'Run Linter'
         run: cargo clippy --all -- -D warnings
 
-  run-test-suite-windows:
-    runs-on: windows-latest
+  run-rust-test-suite:
+    name: 'Run Rust test suite'
+    strategy:
+      matrix:
+        features: ['test-kubo,headers', 'test-kubo,headers,rocksdb']
+        platform: ['ubuntu-latest', 'windows-latest', 'macos-13']
+        toolchain: ['stable']
+        exclude:
+          - platform: 'windows-latest'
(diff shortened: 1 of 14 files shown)


.github/workflows/run_test_suite.yaml
```diff
@@ -78,15 +56,31 @@ jobs:
...
         run: cargo clippy --all -- -D warnings
 
-  run-test-suite-linux:
-    runs-on: ubuntu-latest
+  run-rust-test-suite:
+    name: 'Run Rust test suite'
+    strategy:
+      matrix:
+        features: ['test-kubo,headers', 'test-kubo,headers,rocksdb']
+        platform: ['ubuntu-latest', 'windows-latest', 'macos-13']
```
jsantell, line 64: We use `macos-12` for the swift tests, should these be synced?

jsantell: Also surprised macos doesn't need any dependencies installed

jsantell (approved):
Looking great, this should quiet up some tests while still surfacing issues

.github/workflows/run_test_suite.yaml
cdata: We should update to MacOS 13 if it works.

It's possible that we don't need these dependencies to run the core Rust tests anymore. Or else, that the associated packages already have suitable versions installed on MacOS.

Issue #686, opened 2023-10-19
Updating reqwest to 0.11.22 causes macos build failure, requires SystemConfiguration


jsantell:
In some dev work, I updated `reqwest` from 0.11.20 to 0.11.22 and ran into [failures](https://github.com/subconsciousnetwork/noosphere/actions/runs/6566865808/job/17840254182?pr=549) in a x86_64 macos build:

```
Undefined symbols for architecture x86_64:
  "_SCDynamicStoreCopyProxies", referenced from:
      __ZN20system_configuration13dynamic_store14SCDynamicStore11get_proxies17ha9cee4367b0c1fcbE in libnoosphere.a(system_configuration-46c6672ca6db6ca9.system_configuration.b48ec42a4ca5d16c-cgu.0.rcgu.o)
  "_SCDynamicStoreCreateWithOptions", referenced from:
      __ZN20system_configuration13dynamic_store14SCDynamicStore6create17he63aeeeb3840fdc4E in libnoosphere.a(system_configuration-46c6672ca6db6ca9.system_configuration.b48ec42a4ca5d16c-cgu.0.rcgu.o)
  "_SCNetworkInterfaceCopyAll", referenced from:
      __ZN20system_configuration21network_configuration14get_interfaces17ha41c0e3fd9b2cb6eE in libnoosphere.a(system_configuration-46c6672ca6db6ca9.system_configuration.b48ec42a4ca5d16c-cgu.0.rcgu.o)
  "_SCNetworkInterfaceGetBSDName", referenced from:
      __ZN20system_configuration21network_configuration18SCNetworkInterface8bsd_name17h45a02c56c7f559b2E in libnoosphere.a(system_configuration-46c6672ca6db6ca9.system_configuration.b48ec42a4ca5d16c-cgu.0.rcgu.o)
  "_SCNetworkInterfaceGetInterfaceType", referenced from:
      __ZN20system_configuration21network_configuration18SCNetworkInterface21interface_type_string17hd31e474b334507ecE in libnoosphere.a(system_configuration-46c6672ca6db6ca9.system_configuration.b48ec42a4ca5d16c-cgu.0.rcgu.o)
  "_SCNetworkInterfaceGetLocalizedDisplayName", referenced from:
[...]

PR #688, opened 2023-10-19
chore: Pin reqwest to 0.11.20 to avoid issues with #686.

jsantell:


cdata:
To pin, you need to use the `=` prefix. Otherwise, the default resolution for a version is: `1.2.3  :=  >=1.2.3, <2.0.0`

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio

jsantell:
> To pin, you need to use the `=` prefix. Otherwise, the default resolution for a version is: `1.2.3 := >=1.2.3, <2.0.0`
> 
> https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio

good catch!

cdata (dismissed):
LGTM

cdata (approved):

PR #698, opened 2023-10-25
feat: Initial work refactoring noosphere-gateway to be generic over the spheres it manages.

jsantell:
* `trait ContextManager` is provided to the gateway to access managed spheres' contexts.
* `SingleSphereContextManager` implementation for existing `orb` gateway and tests.
* All gateway routes now require `X-SPHERE-IDENTITY` header to specify the target sphere. Corresponding UCAN auth could apply to many spheres.
* In addition to `GatewayAuthority` synthesized for authenticated routes, there's now `GatewaySphereContext` which provides access to the targeted gateway sphere.
* Use [TypedHeaders](https://docs.rs/headers/latest/headers/index.html) for parsing `X-SPHERE-IDENTITY` and `UCAN` headers into a proof chain. Note using the `headers` module, it should be the same version as `axum` in servers.
* Worker tasks naively loop through all managed sphere contexts before waiting on some timer. Not scalable, and could result in long worker cycles, but good enough for now.

`cargo test --features test-kubo,helpers` passing locally


jsantell:
Updates:
* Requests are now scoped by hostname e.g. `did:key:abc.subconscious.cloud`.
  * Some tests updated to properly set counterpart in gateways
  * `SingleSphereContextManager` always has available its one sphere context, and does not use whatever's in the hostname. This enables existing cluster gateways, and possibly misconfigured self-served gateways to continue functioning. The real counterpart identity is fetched from sphere storage to reflect truth.
  * TBD on what formatting we want for the DID subdomain e.g. `did_key_abc` or elide the DID type altogether `abc.subconscious.cloud`.
* `GatewayAuthority`, in addition to handling auth, now contains the scoped sphere context. For unauthenticated requests (Did route), a `GatewaySphereContext` is used to access the sphere context. Rationale here is to not double up on processing on every request, and use one or the other in a route (with `GatewayAuthority` using an underlying `GatewaySphereContext`.
* `X-SPHERE-IDENTITY` header removed.

jsantell:
Rebased on #685

jsantell:
rebased

cdata:
IIUC what you wrote here correctly:

> GatewayAuthority, in addition to handling auth, now contains the scoped sphere context. For unauthenticated requests (Did route), a GatewaySphereContext is used to access the sphere context. Rationale here is to not double up on processing on every request, and use one or the other in a route (with GatewayAuthority using an underlying GatewaySphereContext.


This is actually a lot more expensive for the DID route, which currently is an unauthorized route that just response with a string: https://github.com/subconsciousnetwork/noosphere/blob/39a58e6335071662fef989935907025d0732e3cf/rust/noosphere-gateway/src/handlers/v0alpha1/did.rs#L5

Note that accessing a `SphereContext` effectively opens a database, and in a stateless world we face doing this on every request. This could lead to DoS by malicious clients hitting the DID route abusively.

cdata:
Can you speak a little bit about what this is building towards? Have you chosen a serverless function stack?

jsantell:
> Can you speak a little bit about what this is building towards? Have you chosen a serverless function stack?

Step 1 is to genericize sphere context access in the gateway to share functionality between single-sphere gateways and a multi-tenant one, using the high-level `start_gateway()` function in `noosphere-gateway`. The minimum interface `ContextManager` supports this currently, with hand waving over `.iter()` for the worker threads. Landing this would let us begin continually integrating MTG progress in our cluster.

Step 2 involves more long term abstraction over how we want to extend the gateway. Using the high level `start_gateway()` doesn't provide extensions to the axum server/custom routing, nor worker configuration, but we can begin to sketch out a shared worker queue for all tasks, compatible with remote MQs, exporting axum routes, etc. -- none of this is needed yet to get a multi-tenant gateway up and running.

No opinions on serverless function stacks; the multi tenant gateway is designed to run as several instances operating over common data. Could become serverless functions if needed. Does not seem immediately relevant to spinning up multi-tenant gateways

> This is actually a lot more expensive for the DID route, which currently is an unauthorized route that just response with a string: 

We can add a more lightweight function in `ContextManager` for resolving the map from counterpart to sphere identity for this route

cdata:
> Does not seem immediately relevant to spinning up multi-tenant gateways

It's relevant insofar as we are writing `Send + Sync` bounds e.g., in `ContextManaager`. If we are generalizing without expressing an opinion at this point, I would like to make sure we aren't ruling out `wasm32-wasi`.

rust/noosphere-gateway/src/context_manager.rs
```diff
@@ -0,0 +1,74 @@
...
+    C: HasMutableSphereContext<S>,
+    S: Storage + 'static,
+{
+    async fn get_sphere_context(&self, _counterpart: &Did) -> Result<C> {
+        // Our original single sphere gateways in the cluster use a hashed identity
+        // in its hostname, so we won't get the correct counterpart here. Similarly,
+        // a self-hosted orb server may not have a proper subdomain for the sphere,
+        // with a gateway already bound to a single sphere (via [SingleSphereContextManager]).
+        // Always return the only context we have. [GatewayAuthority] ensure requests
+        // are properly authorized.
```
cdata, line 65: Some of this rationale belongs in the main doc for `SingleSphereContextManager`. Also, please dance around the implementation details of our cloud offering; it is preferable for such details to be omitted from this open source project.

rust/noosphere-gateway/src/middleware.rs
```diff
@@ -0,0 +1,244 @@
+use anyhow::Result;
```
cdata, line 1: Please make this a folder with an `authority.rs` module inside. I've realized recently that naming the file after what it contains is key to quickly navigating this code base, and I probably won't remember to look for "middleware" when I'm trying to find the "gateway authority."

rust/noosphere-gateway/src/middleware.rs
```diff
@@ -0,0 +1,244 @@
...
+use noosphere_storage::{KeyValueStore, Storage};
+use ucan::{capability::CapabilityView, chain::ProofChain, crypto::did::DidParser};
+
+use crate::ContextManager;
+
+/// Represents the scope of a gateway requests' access to a sphere.
+/// Use this in handlers to access a sphere context without authorizing.
+/// Otherwise, use [GatewayAuthority] which contains authorization
+/// as well as [GatewaySphereContext] functionality.
+pub struct GatewaySphereContext<M, C, S> {
```
cdata, line 24: The naming and implementation here needs some additional thought. Here are my initial reactions:

1. We already have a thing called `SphereContext` and this wrapper is asymmetrical with it
2. The wrapper implementation appears to forward calls to `HasSphereContext` and `HasMutableSphereContext` trait methods, but does not implement those traits itself, which makes me wonder...
3. ...should this wrapper even exist? Can it just be replaced with any `impl HasMutableSphereContext` where it is referenced elsewhere? Could it just be a helper for constructing an `Arc<Mutex<SphereContext<...>>` or similar that is taken from it and passed on to the handlers?

rust/noosphere-gateway/src/middleware.rs
```diff
@@ -0,0 +1,244 @@
...
+/// access.
+///
+/// Embodies the authorization status of the request-maker as it is
+/// represented by its `ucan` headers. Any request handler can use a [GatewayAuthority]
+/// to test if a required capability is satisfied by the authorization
+/// presented by the maker of the request, and provide access to the scoped
+/// sphere context.
+pub struct GatewayAuthority<M, C, S> {
+    proof_chain: ProofChain,
+    sphere: GatewaySphereContext<M, C, S>,
```
cdata, line 121: I'm not sold on the `GatewayAuthority` "owning" the canonical `SphereContext` for the lifetime of the handler. It's not that it doesn't work, just that it seems like an unnecessary conflation of concepts without an obvious benefit.

Collapsing the number of arguments in a handler isn't a clear benefit to me because it obscures what is actually in use by the handler (and since Rust is verbose, it doesn't spare us from all the type bounds needed to express the shape of the `SphereContext`).

If it's about reducing work, I would suggest passing in the `SphereContext` from elsewhere at the time it is needed.

Otherwise, perhaps there is a join construct that pairs a `GatewayAuthority` with a `SphereContext`.

Are there other benefits to merging these concepts that I might be missing?

rust/noosphere-gateway/src/middleware.rs
```diff
@@ -0,0 +1,244 @@
...
+
+fn map_bad_request<E: std::fmt::Debug>(error: E) -> StatusCode {
+    error!("{:?}", error);
+    StatusCode::BAD_REQUEST
+}
+
+/// Takes arguments from a [FromRequestParts::from_request_parts] call and parses out
+/// the counterpart sphere identity from hostname.
+/// e.g. `did_key_abc.subconscious.cloud` -> `Did(String::from("did_key_abc"))`
+async fn get_counterpart_from_hostname<State>(
```
cdata, line 225: This needs some careful thought, because pulling the counterpart out of the hostname is not something we want to formalize as a "Noosphere Protocol" detail. I would be more comfortable if a "counterpart resolution" function was given to the gateway at the time it is instantiated.

rust/noosphere/tests/client_to_gateway.rs
```diff
@@ -55,22 +51,24 @@ async fn gateway_tells_you_its_identity() -> Result<()> {
...
     let client_sphere_identity = client_workspace.sphere_identity().await.unwrap();
-
-    let gateway_sphere_context = gateway_workspace.sphere_context().await.unwrap();
+    let mut gateway_sphere_context = gateway_workspace.sphere_context().await.unwrap();
+    {
+        let mut sphere_context = gateway_sphere_context.sphere_context_mut().await?;
+        sphere_context
+            .db_mut()
+            .set_key(COUNTERPART, client_sphere_identity)
+            .await?;
```
cdata, line 61: In the spirit of these tests, lets use the CLI command implementation e.g., `config_set(ConfigSetCommand::Counterpart { did }, &mut gateway_workspace)`

https://github.com/subconsciousnetwork/noosphere/blob/39a58e6335071662fef989935907025d0732e3cf/rust/noosphere-cli/src/native/commands/sphere/config.rs#L11-L24

rust/noosphere/tests/client_to_gateway.rs
```diff
@@ -119,22 +117,24 @@ async fn gateway_identity_can_be_verified_by_the_client_of_its_owner() -> Result
...
```
cdata, line 127: Same comment as above

rust/noosphere/tests/client_to_gateway.rs
```diff
@@ -191,22 +191,24 @@ async fn gateway_receives_a_newly_initialized_sphere_from_the_client() -> Result
...
```
cdata, line 201: Ditto

rust/noosphere/tests/client_to_gateway.rs
```diff
@@ -283,22 +285,24 @@ async fn gateway_updates_an_existing_sphere_with_changes_from_the_client() -> Re
...
```
cdata, line 295: Etc.

cdata (changes requested):
This is a good first pass. I think some aspects need a bit of thought, but I'm excited about the direction it's going in.

rust/noosphere-cli/src/native/helpers/workspace.rs
```diff
@@ -43,7 +40,7 @@ pub fn temporary_workspace() -> Result<(Workspace, (TempDir, TempDir))> {
 }
 async fn start_gateway_for_workspace(
     workspace: &Workspace,
-    client_sphere_identity: &Did,
+    _client_sphere_identity: &Did,
```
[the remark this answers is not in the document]
jsantell: If this design looks good, yup can be removed

rust/noosphere-core/src/api/headers/ucan.rs
```diff
@@ -0,0 +1,90 @@
...
+        for (cid, jwt) in self.0.iter() {
+            let value = HeaderValue::from_str(&format!("{} {}", cid, jwt)).unwrap();
+            values.extend(std::iter::once(value));
+        }
+    }
+}
+
+impl Ucan {
+    /// Construct a [ProofChain] from the collected `ucan` header
+    /// values. This is where validation of the data occurs.
```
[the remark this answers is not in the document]
jsantell: No logic changes here if this `Header` implementation correctly parses *all* ucan headers before handling here (hence the `Vec<(Cid, Jwt)>`), which all get written to the store before proof chain is constructed, IIUC

rust/noosphere-gateway/src/gateway.rs
```diff
@@ -79,64 +65,64 @@ where
...
+
+    It has bound a gateway to {:?}
+    It awaits updates from sphere {}..."#,
+            gateway_scope.identity,
+            listener
+                .local_addr()
+                .expect("Unexpected missing listener address"),
+            gateway_scope.counterpart
+        );
+        */
```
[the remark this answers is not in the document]
jsantell: Possibly:
* ContextManager calls a success callback on successful server start that can print out a message
* Return a task or wrapper from `start_gateway()` that needs to be awaited to actually start the server, giving consumer a moment to log whatever

Neither feels correct compared to exposing routes and a (new) worker abstraction, such that the STG is the only caller of `start_gateway` (and we can then leave it in), but maybe a ContextManager callback is the quickest path forward

rust/noosphere-gateway/src/handlers/v0alpha1/did.rs
```diff
@@ -1,6 +1,22 @@
...
+use crate::{middleware::GatewaySphereContext, ContextManager};
+
+pub async fn did_route<M, C, S>(sphere: GatewaySphereContext<M, C, S>) -> Result<String, StatusCode>
+where
+    M: ContextManager<C, S>,
+    C: HasMutableSphereContext<S>,
+    S: Storage + 'static,
+{
+    let did = sphere
+        .sphere_context()
```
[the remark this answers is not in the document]
jsantell: A lightweight `get_sphere_id(counterpart: &Did)` or something may fit the bill here. Database (postgres) access is inevitable, though we don't need to create a sphere context. We've worked around the "new database connection on request" as its a non-starter for pretty much all DBs

rust/noosphere-gateway/src/middleware.rs
jsantell: Goal here is to reduce work -- IIRC, `FromRequestParts` is processed unordered, and if we need both authority and sphere context in a route, and the authority always needs a sphere context, then lets combine them. I agree that there's some conflating ideas here. There are a handful of ways of processing requests in axum, so maybe there's another one that fits the bill -- `Layers`, for example, is executed in order and does not take ownership IIRC. With functionality working currently, I can explore other options here (like `Layer`)

rust/noosphere-gateway/src/middleware.rs
jsantell: My understanding was that this *is* a part of the protocol, due to secure origins on the web. Exposing this (currently, without a more modular V2 that exposes individual routes where a consumer BYO's axum) with axum guts may get weird. WDYT if we had two options for counterpart parsing, one as a subdomain, one as a header or query param, not dissimilar from IPFS gateways, such that the implementations live with the other axum stuff for now?

rust/noosphere-core/src/api/headers/ucan.rs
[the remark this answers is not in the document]
jsantell: Added spec link and a note that we are not currently adhering to the 0.3.0 spec with a single `ucans` header, rather than multiple `ucan` headers. #708

rust/noosphere-gateway/src/middleware.rs
cdata: > My understanding was that this is a part of the protocol, due to secure origins on the web

Nope, this is a requirement for our cloud if we intend to serve a client over the web, because we'll be serving multiple customers and we don't want their origins to cross over.

A self-hosting individual user doesn't have the same requirement.

rust/noosphere-gateway/src/middleware.rs
cdata: Again, it's not that it doesn't work. My complaint is about the conceptual flattening happening here. The authority and the sphere context are two different concepts. If we were to make the claim that they are always delivered together, then I would suggest that there be a super construct to join them.

rust/noosphere-gateway/src/handlers/v0alpha1/did.rs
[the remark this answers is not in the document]
cdata: We're talking about a static string that is always the same, with a 1:1 mapping to the input (in the cloud case, the value extracted from the `Host` header). We will not need to open a connect to the DB to serve it in a response.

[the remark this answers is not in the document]
cdata: Here are some docs on GCP Cloud CDN: https://cloud.google.com/cdn/docs/caching

It should be possible for us to craft a response that can be served by Cloud CDN for most requests. There are probably some other optimizations to consider as well.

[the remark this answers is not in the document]
cdata: And just for the sake of observing the balance that needs to be struck here: serving a static string allows us to be "optimal" in both the self-host and cloud case. A self-hosting user cannot be assumed to be running a sophisticated caching layer in front of their node, so we should strive to be efficient in that case where reasonable.

rust/noosphere-gateway/src/gateway.rs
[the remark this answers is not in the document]
cdata: I think what we need to do is extract all the Axum setup stuff, so that you can get an `app` without any superfluous logging and then invoke `Server::from_tcp(listener)?.serve(app.into_make_service())` separately.


Then, we can have two different `start_gateway`-like  entrypoints: one that may be implemented for the CLI, and one that may be implemented for cloud use.

rust/noosphere-gateway/src/handlers/v0alpha1/did.rs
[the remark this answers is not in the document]
jsantell: * In all instances, gateways will only ever **open** up a single database connection (or e.g. a single connection pool as an implementation detail)
* Setting some cache headers sounds like an excellent optimization, especially if we only need to set some headers from the server. Is this needed for a V1 here?
* In lieu of a cache or cache expiry, are you suggesting that this can resolve *without* a (postgres, no block storage) query?

rust/noosphere-gateway/src/middleware.rs
jsantell: This was cleaned up significantly in the latest push, implementing directly `HasSphereContext` and `HasMutableSphereContext`. We can't implement `FromRequestParts` for e.g. `C`, hence the newtype (although technically not a new type because there's PhantomData)

rust/noosphere-gateway/src/middleware.rs
jsantell: Agreed on all fronts, new change addresses these well I think

rust/noosphere-gateway/src/middleware.rs
jsantell: Addressed!

rust/noosphere-gateway/src/gateway.rs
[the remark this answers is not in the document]
jsantell: Implemented via

```rust
let gateway = Gateway::new(manager, ipfs_url, ns_url, None)?;
// log here
gateway.start(listener).await?;
```

rust/noosphere-cli/src/native/commands/serve.rs
```diff
@@ -19,25 +20,30 @@ pub async fn serve(
...
-    .await
+    let gateway_identity = sphere_context
+        .sphere_context()
+        .await?
+        .author()
+        .did()
+        .await?;
+    let manager = SingleTenantGatewayManager::new(sphere_context, counterpart.clone()).await?;
+
+    let gateway = Gateway::new(manager, ipfs_api, name_resolver_api, cors_origin)?;
```
cdata, line 33: :+1:

rust/noosphere-gateway/src/handlers/v0alpha1/push.rs
```diff
@@ -50,23 +46,25 @@ where
...
+        .try_authorize(
+            &mut gateway_sphere,
+            counterpart,
+            &generate_capability(counterpart.as_str(), SphereAbility::Push),
+        )
+        .await?;
 
     let gateway_push_routine = GatewayPushRoutine {
-        sphere_context,
+        gateway_sphere,
```
cdata, line 63: This naming tripped me up in my last change (and I'm the original author for crying out loud), so thanks for clarifying it!

rust/noosphere-gateway/src/handlers/v0alpha1/push.rs
```diff
@@ -325,7 +317,7 @@ where
     async fn notify_name_resolver(&self) -> Result<()> {
         if let Some(name_record) = &self.request_body.name_record {
             if let Err(error) = self.name_system_tx.send(NameSystemJob::Publish {
-                context: self.sphere_context.clone(),
+                context: self.gateway_sphere.to_inner(),
```
cdata, line 320: Why isn't it sufficient to pass the `GatewaySphereContext` directly here? That is to ask: why `to_inner`?

rust/noosphere-gateway/src/handlers/v0alpha2/push.rs
```diff
@@ -149,7 +153,8 @@ where
             return Err(PushError::UnexpectedBody);
         };
 
-        let gateway_sphere_tip = self.sphere_context.version().await?;
+        let gateway_sphere_context = self.gateway_sphere.sphere_context().await?;
+        let gateway_sphere_tip = gateway_sphere_context.version().await?;
```
cdata, line 157: Why not `self.gateway_sphere.version()`?

rust/noosphere-gateway/src/single_tenant.rs
```diff
@@ -0,0 +1,81 @@
...
+            true => Ok(self.identity.clone()),
+            false => Err(anyhow!(
+                "No sphere identity found for counterpart: {counterpart}."
+            )),
+        }
+    }
+
+    fn iter(&self) -> Pin<Box<GatewayManagerSphereStream<'_, C>>> {
+        Box::pin(try_stream! {
+            yield self.context.clone();
```
cdata, line 74: Consider `tokio_stream::once`: https://docs.rs/tokio-stream/latest/tokio_stream/fn.once.html

cdata (commented):
Tentatively this looks like a candidate for merge, but let's talk F2F about some of my last comments!

rust/noosphere-gateway/src/gateway.rs
```diff
@@ -3,148 +3,137 @@ use axum::extract::DefaultBodyLimit;
...
-    ipfs_api: Url,
-    name_resolver_api: Url,
-    cors_origin: Option<Url>,
-) -> Result<()>
-where
-    C: HasMutableSphereContext<S> + 'static,
-    S: Storage + 'static,
-{
-    initialize_tracing(None);
+type WorkerHandles = Vec<JoinHandle<Result<()>>>;
```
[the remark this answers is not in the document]
jsantell: Aye more specifically, will have to be generic over embedded workers (current impl, either multi or single) vs remote workers

rust/noosphere-gateway/src/gateway_manager.rs
```diff
@@ -0,0 +1,39 @@
...
+    /// Retrieve a sphere context whose counterpart matches `counterpart`.
+    async fn get_sphere_context(&self, counterpart: &Did) -> Result<C>;
+
+    /// Retrieve the identity of the managed sphere's device key from provided `counterpart`,
+    /// as a lightweight alternative to fetching the entire sphere context
+    /// in [GatewayManager::get_sphere_context].
+    async fn get_gateway_identity(&self, counterpart: &Did) -> Result<Did>;
+
+    /// Iterate over all managed spheres. Currently used by task workers.
+    /// Will need to rethink this.
```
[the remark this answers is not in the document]
jsantell: Aye the GatewayManager should never be accessed under a route handler, outside of the middleware parsing which pulls out the relevant scope information. The `iter()` function is only used as a stop-gap for current embedded workers, and is conflated with `GatewayWorker` currently for migration ease. This should go away once migrating to a new vision of workers -- will clarify those plans in a follow up issue.

In the mean time, there may be some Axumisms we can use to prevent pulling this in as a router arg, or some scary renaming e.g. `.experimental_worker_only_iter()`

rust/noosphere-gateway/src/handlers/v0alpha1/push.rs
jsantell: `C` is always the inner sphere context -- `Arc<Mutex<SphereContext<S>>>` I believe. Returning the `gateway_sphere` is a different newtype. Unfortunately, newtype is required since we can't have an extractor for a generic `C` type

rust/noosphere-gateway/src/handlers/v0alpha1/push.rs
jsantell: hah, the spheres and contexts trip me up too

rust/noosphere-gateway/src/handlers/v0alpha2/push.rs
jsantell: The proceeding `sphere_context()` locks the mutex (hanging the `gateway_sphere.version()` call) -- getting the version from the `gateway_sphere` first would just incur an extra lock, IIUC

rust/noosphere-gateway/src/handlers/v0alpha1/push.rs
cdata: This just begs the question: why can't `C` be the `GatewaySphereContext` type?

jsantell: Documenting different approaches here on matching the `C` type with the arguments used in routes:

* Approach 1: `C: HasMutableSphereContext<S> + FromRequestParts<_, _> `: we can't generally implement FRP for C (S not trait bound), having the consumer implement is fraught due to the trait being dependent on axum state/middleware. We could implement for concrete types of C maybe (`Arc<Mutex<SphereContext<S>>>`), but then our top level generics aren't that generic. thinking more, i think we explicitly do want to implement FRP over a newtype, as its what interacts with the gateway manager, all specific to our other extractors
* Approach 2: `Gateway::new::<M, GatewaySphereContext<Arc<Mutex<SphereContext<S>>>>, S>(..)`  we could expose our newtype wrapper, with its own HasMutableSphereContext traits, but seems irrelevant on the consumer side
* Approach 3: `sphere_extractor: SphereExtractor<C, S>` implement FRP over a C wrapper and immediately convert to C; not elegant though not much different from Arc<T> types

opting for 3 here, as for now we explicitly want to define the FRP used in routes, and while it'd be nice for routes to have `route_handler(context: C)`, introducing a newtype only for axum parsing seems reasonable

rust/noosphere-gateway/src/extractors/scope.rs
```diff
@@ -0,0 +1,57 @@
...
+/// Represents the scope of a gateway request as a counterpart [Did],
+/// and the corresponding managed sphere's author/device key,
+/// the gateway identity.
+///
+/// Extracting a [GatewayScope] is efficient, and does not open
+/// a [SphereContext].
+pub struct GatewayScope<C, S> {
+    pub counterpart: Did,
+    pub gateway_identity: Did,
+    c_marker: PhantomData<C>,
```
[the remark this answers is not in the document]
cdata: I think this wasn't actually resolved?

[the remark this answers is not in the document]
jsantell: Didn't save in VSCode :cry: pushed changes

cdata (approved):
Great work on this!

PR #703, opened 2023-10-31
fix: Recovery only uses latest version of sphere

cdata:
This change modifies the `/v0alpha1/replicate` route to accept a `include_content` parameter. The use-case of the parameter is to facilitate a recovery via replication.

Additionally, recovery has been changed to use `/v0alpha1/replicate` instead of `/v0alpha1/fetch`. We are doing this to head-off the scaling fault where a user cannot successfully recovery if their gateway's sphere has a really long history.

The most notable consequence of this change is that a user who recovers their sphere from a gateway will not have any of the "history" of changes prior to the recovery point available on the recovery device. We'll have to reckon with this choice at some point down the road when we move to support end-user views into history.

As of this change, `v0alpha1/fetch` and `/v0alpha1/replicate` are probably redundant (with `fetch` being a subset of `replicate`), but for now I think we should leave both in place and decide how to consolidate them in a future change.


cdata:
Haven't seen this one flake before:

>    FLAKY 2/6 [   1.896s] noosphere-ns dht::node::test::test_dhtnode_bootstrap

jsantell:
> Haven't seen this one flake before:
> 
> > FLAKY 2/6 [   1.896s] noosphere-ns dht::node::test::test_dhtnode_bootstrap

I've seen this in another test, ~~didn't expect a panic~~ looks like `await_or_timeout()` function panics after 700ms of not achieving its success state. These tests were always a bit racey, could be improved here

```
--- TRY 1 STDERR:        noosphere-ns dht::node::test::test_dhtnode_bootstrap ---
thread 'dht::node::test::test_dhtnode_bootstrap' panicked at rust/noosphere-ns/src/dht/node.rs:249:42:
timed out: waiting for 4 peers
```

jsantell (approved):
Looking great :+1:

PR #719, opened 2023-11-09
chore: increase timeout on DHT node initialization

jsantell:


cdata (approved):
LGTM - a rationale would be nice in future PR descriptions.

PR #726, opened 2023-11-20
feat: Rewrite host headers in API client to embed sphere identity in subdomain

jsantell:
...when feature 'test-gateway' is enabled.

### DIFF
--- a/rust/noosphere-core/src/api/client.rs
+++ b/rust/noosphere-core/src/api/client.rs
@@ -21,6 +21,7 @@ use noosphere_storage::{block_deserialize, block_serialize, BlockStore};
 use reqwest::{header::HeaderMap, StatusCode};
 use tokio_stream::{Stream, StreamExt};
 use tokio_util::io::StreamReader;
+use tracing::*;
 use ucan::{
     builder::UcanBuilder,
     capability::CapabilityView,
@@ -30,7 +31,7 @@ use ucan::{
 };
 use url::Url;
 
-#[cfg(doc)]
+#[cfg(any(doc, feature = "test-gateway"))]
 use crate::data::Did;
 
 use super::v0alpha1::ReplicationMode;
@@ -39,6 +40,10 @@ use super::v0alpha1::ReplicationMode;
 /// API. It embodies the intended usage of the REST API, which includes an
 /// opening handshake (with associated key verification) and various
 /// UCAN-authorized verbs over sphere data.
+///
+/// When built with the `test-gateway` flag, a modified Host header
+/// is sent, rewriting Host to `{IDENTITY}.gateway.test`, where IDENTITY
+/// is the client sphere identity with the `did:key:` prefix stripped.
 pub struct Client<K, S>
 where
     K: KeyMaterial + Clone + 'static,
@@ -61,6 +66,9 @@ where
     pub store: S,
 
     client: reqwest::Client,
+
+    #[cfg(feature = "test-gateway")]
+    forced_host_header: reqwest::header::HeaderValue,
 }
 
 impl<K, S> Client<K, S>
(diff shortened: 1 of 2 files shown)


rust/noosphere-core/src/api/client.rs
```diff
@@ -529,3 +567,20 @@ where
...
+fn apply_test_header(headers: &mut HeaderMap, forced_host_header: &reqwest::header::HeaderValue) {
+    use reqwest::header::HOST;
+    _ = headers.remove(HOST);
+    headers.insert(HOST, forced_host_header.to_owned());
+}
+
+#[cfg(feature = "test-gateway")]
+fn create_test_header(identity: &str) -> Result<reqwest::header::HeaderValue> {
+    Ok(reqwest::header::HeaderValue::from_str(&format!(
+        "{}.gateway.test",
```
cdata, line 581: Would it be possible scope the subdomain to the host of the configured gateway URL e.g., if user configured `http://localhost:8493` as their gateway URL, the host becomes `$SPHERE_ID.localhost`?

rust/noosphere-core/src/api/client.rs
```diff
@@ -529,3 +567,56 @@ where
...
+    Ok(reqwest::header::HeaderValue::from_str(&new_host)?)
+}
+
+#[cfg(all(test, feature = "test-gateway"))]
+mod tests {
+    use super::*;
+    use reqwest::header::HeaderValue;
+
+    #[test]
+    fn test_create_test_header() -> Result<()> {
```
cdata, line 601: Preferred naming for tests is `it_$description_of_thing_it_does`

rust/noosphere-core/src/api/client.rs
```diff
@@ -529,3 +567,56 @@ where
...
+
+#[cfg(feature = "test-gateway")]
+fn apply_test_header(headers: &mut HeaderMap, forced_host_header: &reqwest::header::HeaderValue) {
+    use reqwest::header::HOST;
+    _ = headers.remove(HOST);
+    headers.insert(HOST, forced_host_header.to_owned());
+}
+
+#[cfg(feature = "test-gateway")]
+fn create_test_header(api_base: &Url, identity: &str) -> Result<reqwest::header::HeaderValue> {
```
cdata, line 579: Use `&Did` instead of `&str` here

rust/noosphere-core/src/api/client.rs
```diff
@@ -529,3 +567,56 @@ where
...
+    let mod_identity = identity
+        .strip_prefix("did:key:")
+        .ok_or_else(|| anyhow!("Could not format HOST header for test-gateway."))?;
```
cdata, line 582: `Host` or `host`, not `HOST` https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host

rust/noosphere-core/src/api/client.rs
```diff
@@ -529,3 +567,56 @@ where
...
+    let domain = api_base.domain().or_else(|| Some("gateway.test")).unwrap();
```
cdata, line 583: Probably want this to be an error, since it is just guessing at this point?

cdata (commented):
This is great, minor cleanup requested!

rust/noosphere-core/src/api/client.rs
jsantell: domain is `None` in the case of e.g. `127.0.0.1`, where we need some domain name to use to prepend the sphere identity

cdata: Yeah, that's why I'm saying it should be an error, because you're just guessing that e.g., `gateway.test` is a legible hostname.

cdata: If the tester explicitly wants that hostname, they can configure it!

jsantell: I'm not sure I follow, this is strictly to test servers at host 127.0.0.1 in tests, if I could configure a host name for local tests this feature wouldn't be needed. `gateway.test` can be anything, but we can't add a subdomain to `127.0.0.1`

jsantell:
Updated, rejecting if no domain found; downstream usage requires e.g. `localhost` for `127.0.0.1`, but can support a subdomain for counterpart extraction (updated tests)

cdata (approved):
Thanks for taking the feedback in stride. LGTM!

PR #734, opened 2023-11-27
feat: Force quit rocksdb worker threads upon dropping.

jsantell:


cdata:
@jsantell can you provide some commentary on the motivation of this change? Most of the context is phased out of my brain from holidays.

jsantell:
Suspecting a possible culprit for the lock inbetween windowing: https://github.com/subconsciousnetwork/subconscious/issues/1008 -- not sure of the current state there, but this at least reasonably maps rocksdb resource lifetimes to the rust structures (originally elided due to stack overflow in early integration tests and a one off comment somewhere suspecting the two related)

cdata (approved):
LGTM

PR #767, opened 2024-01-08
fix!: The gateway should not call initialize_tracing

justinabrahms:
Initializing tracing is the responsibility of a surrounding binary, not library code.


jsantell (dismissed):
Aye `initialize_tracing` should only be called in `main` functions in executables and test functions; running a gateway (via `orb`) initializes this here: https://github.com/subconsciousnetwork/noosphere/blob/5a84d7a427d01d649ebfa814011bad5a8f5162db/rust/noosphere-cli/src/native/mod.rs#L45

jsantell (approved):

PR #779, opened 2024-01-09
feat: Add support for BYO tracing_subscriber

justinabrahms:
Allows exfiltration of instrumentation from noosphere into handy places like otel syncs like:

```rust
    let tracer = opentelemetry_otlp::new_pipeline()
        .tracing()
        .with_trace_config(
            trace::config().with_resource(Resource::new(vec![KeyValue::new(
                "service.name",
                "my service",
            )])),
        )
        .with_exporter(opentelemetry_otlp::new_exporter().tonic())
        .install_batch(opentelemetry_sdk::runtime::Tokio)?;

    let otel = tracing_opentelemetry::layer().with_tracer(tracer);

    initialize_tracing_with_layer(None, "multi-tenant gateway", otel);

```

### DIFF
--- a/rust/noosphere-core/src/tracing.rs
+++ b/rust/noosphere-core/src/tracing.rs
@@ -256,6 +256,8 @@ mod inner {
         });
     }
 
+    /// Identical to [initialize_tracing], but provides the ability to add in
+    /// your own [Layer] for tracing.
     pub fn initialize_tracing_with_layer<T>(noosphere_log: Option<NoosphereLog>, layer: T)
     where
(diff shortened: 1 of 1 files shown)


cdata:
This looks like a good step in the right direction, although I still think we want a pattern where Noosphere-specific tracing behavior is encapsulated as its own `Layer`, and configuration of the `Subscriber` happens _outside_ of this code.

jsantell (dismissed):
With the wrangling of the multiple layer types, this is a nice & simple way of extending functionality. LGTM with lint passing!

jsantell (approved):

Issue #753, opened 2023-12-12
GatewayManager should conservatively use `extract_counterpart()`


jsantell:
Currently, our gateway extractors extract `GatewayScope`, `SphereExtractor` and `GatewayAuthority`. We should be conservative in calling `extract_counterpart` on unauthenticated requests, or before authentication in authenticated requests, as this could involve database lookups or other external resolution measures.

For authenticated routes, we extract the counterpart to get the sphere context and scope before authenticating within the route. We should explore some shallow UCAN validation before extracting the counterpart. We'll most likely have to rework axum's layers.

For unauthenticated routes (did route), we'll have to resolve the counterpart ultimately. A caching layer here looks appropriate, as the response should very rarely change.

PR #796, opened 2024-01-24
chore: Task Workers Revamp

jsantell:
### Decoupled Worker Tasks

Worker tasks are now abstracted such that routes now send jobs via a `JobClient` decoupled from where the work is performed. A standalone `noosphere_gateway::jobs::worker_queue` module provides the work distribution logic, with each processor receiving a `GatewayJobContext` that provides access to sphere contexts, IPFS, and name resolvers when processing a job.

* Worker jobs can now queue up additional tasks on success (IPFS Syndication -> Name Record Publish) (https://github.com/subconsciousnetwork/noosphere/issues/606).
* Jobs can be retried upon failure. 
* Some timeout logic has been implemented, but not yet ready.
* Republishing name records is now its own job type. Eliminates the need for the scheduler to resolve the latest name record.

### Minimize external queries

`GatewayManager` and routes were updated to minimize GM queries (https://github.com/subconsciousnetwork/noosphere/issues/753).

In addition to a single, cacheable baseline query for scope, unauthenticated routes now has one less external query, and authenticated routes also has one less external query, with another query moved behind authorization.

Ultimately handler logic *should not* be changed, other than moving Push handler's queueing of name record publishing behind IPFS syndication.

## Caveats/Concerns/Enhancements

* We should do some profiling to measure lock contention across workers where appropriate amongst shared resources like `KuboClient` and the name resolver.
* Job processors may need to be rethought in order to better take advantage of retries e.g. classifying a failure between retriables, invalid state/unretriable, and failed-but-ok-to-lazily-try-again-later.
* ~~Implement timeouts in worker queue.~~ Timeouts now implemented
* Ipfs syndication revision parameter now optional (unused? to confirm)
* Ensure name record provided to PUSH is still valid after Ipfs syndication completes due to record expiry


jsantell:
Added explicit checks on existence of name records after syncing changes in tests, possibly clearing up a handful of intermittent failures.

Windows builds are having an unrelated issue: https://github.com/subconsciousnetwork/noosphere/pull/805

rust/noosphere-gateway/src/single_tenant/job_client.rs
```diff
@@ -0,0 +1,168 @@
...
+/// The number of worker threads to process jobs.
+const WORKER_COUNT: usize = 5;
+/// How many times by default jobs should retry upon failure.
+/// @TODO Consider allowing individual jobs to hint if they
+/// should be retried, versus failures that can be attempted
+/// on next periodic cycle.
+const JOB_RETRIES: usize = 1;
+/// How many seconds before a job is considered broken,
+/// and potentially restarted.
+const TIMEOUT_SECONDS: u64 = 180;
```
jsantell, line 30: Syndicating large histories could be longer than this timeout, and notably, name records' publishing timeframe is currently 2 minutes.

rust/noosphere-gateway/src/jobs/worker_queue/queue.rs
```diff
@@ -0,0 +1,70 @@
...
+    /// By default, `retries` is set to 1 and `timeout` is 3 minutes.
+    pub fn spawn(
+        worker_count: usize,
+        worker_context: P::Context,
+        retries: Option<usize>,
+        timeout: Option<Duration>,
+    ) -> Self {
+        let (request_tx, request_rx) = unbounded_channel();
+        let handle = Some(Arc::new(tokio::spawn(async move {
+            let workers = WorkerQueueThread::<P>::new(
```
cdata, line 35: The naming and documentation were confusing to read the first time around. `WorkerQueue` sounds like "the thing that has a pool of workers." In fact, the documentation currently reads:

```rs
/// An abstraction around managing several worker threads, and distributing
/// work amongst them.
```

And, there is `WorkerQueueThread`, which sounds like the thing that does work on a thread. But, most of the meaningful management of multiple threads is performed by `WorkerQueueThread`.

Why are these separate constructs? It kind of looks like all that `WorkerQueue` does is thinly wrap `WorkerQueueThread`. Also: can we call `WorkerQueueThread` something closer to what it is? Maybe it's just `WorkerQueue`, maybe it's a `WorkerQueuePool`. It's certainly not a single thread so long as it has a `Vec<Workers<P>>`.

rust/noosphere-gateway/src/jobs/worker_queue/worker.rs
```diff
@@ -0,0 +1,151 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("JobRequest")
+            .field("job", &self.job)
+            .field("attempts", &self.attempt)
+            .finish()
+    }
+}
+
+/// Provides an interface to perform a job via [Processor]
+/// on another thread.
```
cdata, line 54: Can you comment about the usage and lifecycle of a `Worker` in the documentation?

rust/noosphere/tests/distributed_basic.rs
```diff
@@ -292,35 +295,45 @@ async fn traverse_spheres_and_get_incremental_updates_via_noosphere_gateway_via_
...
     for pair in [&pair_1, &pair_2] {
         let name = pair.name.clone();
         let mut ctx = pair.sphere_context().await?;
         ctx.write("my-name", "text/plain", name.as_ref(), None)
             .await?;
         ctx.save(None).await?;
-        ctx.sync().await?;
+        initial_versions.push(ctx.sync().await?);
+    }
+    {
```
cdata, line 319: Why is the extra block needed?

cdata (changes requested):
The bones look excellent, but I have some light opinions about some of the flesh :)

Great work so far!

rust/noosphere/tests/distributed_basic.rs
jsantell: Looks unnecessary, will remove

jsantell:
Updates ([commit](https://github.com/subconsciousnetwork/noosphere/pull/796/commits/8cadcfc18fb9e514ebf7d4426bd97bda06476e11)):
* Changes/cleanups from feedback
* Remove vestigial `revision` in IpfsSyndication job
* Reduce indirection from cloning (e.g. use outer Arcs, not inner)
* s/ContextResolver/SphereContextResolver
* s/NameResolverExt/NameResolverPoller
* s/WorkerQueueThread/WorkerQueueOrchestrator
* More documentation for `worker_queue`

cdata (approved):
LGTM, thanks for the hard work on this!

PR #812, opened 2024-02-08
chore: Update libp2p to 0.53.2

jsantell:
One major need for this is to continue building on rust nightly 1.78.0, due to our current libp2p version using an incompatible crate in its dependency graph.

Docker images are also updated to rust stable (1.75.0) to support dependency updates.

### DIFF
--- a/rust/noosphere-ns/src/dht/swarm.rs
+++ b/rust/noosphere-ns/src/dht/swarm.rs
@@ -1,60 +1,59 @@
-use crate::dht::errors::DhtError;
-use crate::dht::DhtConfig;
+use crate::dht::{DhtConfig, DhtError};
+use anyhow::anyhow;
 use libp2p::{
     allow_block_list,
-    core::muxing::StreamMuxerBox,
-    core::transport::Boxed,
-    core::upgrade,
-    dns,
     identify::{Behaviour as Identify, Config as IdentifyConfig, Event as IdentifyEvent},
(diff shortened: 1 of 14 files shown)


rust/noosphere-ns/src/dht/swarm.rs
```diff
@@ -83,8 +82,20 @@ impl DhtBehavior {
...
+                KademliaBehaviour::with_config(local_peer_id.to_owned(), store, cfg);
+
+            // Updating to libp2p 0.53 introduced DHT nodes adjusting their modes dynamically,
+            // where if a node does not have an external address, it switches to `Mode::Client`.
+            // This improves the network by culling non-accessible nodes.
+            // Manually setting to `Mode::Server` mode prevents this dynamic switching.
+            // We could explore this optimization in the future, but for now,
+            // we need to run on nightly rust, and some challenges with external addresses
+            // in tests.
+            // https://github.com/libp2p/rust-libp2p/pull/4503
```
cdata, line 96: Can we file an issue for "explore this optimization in the future"?

cdata (dismissed):
LGTM, please file requested issue before merge!

rust/noosphere-ns/src/dht/swarm.rs
jsantell: #814

cdata (approved):

PR #818, opened 2024-02-12
fix: Remove vestigial getrandom dep

cdata:
Notes:
- `wasm-bindgen` deprecates `serde-serialize`: https://github.com/rustwasm/wasm-bindgen/issues/2770#issuecomment-1213637387


jsantell:
Handling in #850

PR #819, opened 2024-02-12
[WIP] Cli examples

jsantell:
Examples of some dev scripts in bash and python


scripts/noodev-build
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
```
jsantell, line 1: This is a dupe of noodev-install, ignore

PR #840, opened 2024-02-28
feat: Roll noosphere-ucan-key-support into noosphere-ucan, replace ed25519-zebra with ed25519-dalek.

jsantell:
* Roll `noosphere-ucan-key-support` into `noosphere-ucan` under feature flags: `ed25519`, `rsa`, `p256` and `web-rsa`
  * `ed25519` enabled (indirectly) in tests, replacing the outdated `did-key` (pulling in an old version of [ed25519-dalek](https://github.com/subconsciousnetwork/noosphere/security/dependabot/24) with a vuln, though used only in tests).
  * Add all features in CI tests
* Replace `ed25519-zebra` with `ed25519-dalek`.
* Replace `log` dependency with `tracing`.


rust/noosphere-ucan/Cargo.toml
```diff
@@ -19,6 +19,10 @@ rust-version = "1.75.0"
 
 [features]
 default = []
+ed25519 = ["dep:ed25519-dalek"]
+rsa = ["dep:rsa", "dep:sha2"]
+p256 = ["dep:p256"]
+web-rsa = ["dep:rsa", "dep:sha2", "dep:js-sys", "dep:web-sys", "dep:wasm-bindgen", "dep:wasm-bindgen-futures"]
```
jsantell, line 25: Maybe should be `web-crypto-rsa`?

rust/noosphere-ucan/Cargo.toml
```diff
@@ -31,23 +35,39 @@ futures = { workspace = true }
...
+ed25519-dalek = { workspace = true, features = ["rand_core"] }
+
+[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
+tokio = { workspace = true, features = ["macros", "rt", "test-util"] }
+
+[target.'cfg(target_arch = "wasm32")'.dev-dependencies]
+pollster = { version = "0.3.0" }
+
+[build-dependencies]
+npm_rs = "1.0"
```
jsantell, line 73: I'm not sure if/how this is used

rust/noosphere-ucan/Cargo.toml
```diff
@@ -31,23 +35,39 @@ futures = { workspace = true }
...
 serde_ipld_dagcbor = { workspace = true }
 wasm-bindgen-test = { workspace = true }
+# Always include ed25519 in tests
+ed25519-dalek = { workspace = true, features = ["rand_core"] }
+
+[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
+tokio = { workspace = true, features = ["macros", "rt", "test-util"] }
+
+[target.'cfg(target_arch = "wasm32")'.dev-dependencies]
+pollster = { version = "0.3.0" }
```
jsantell, line 70: I don't think this is used (from ucan_key_support)

cdata (dismissed):
LGTM

rust/noosphere-ucan/Cargo.toml
cdata: Its use is for running async work synchronously in wasm, but I don't think we are doing that anywhere in here at this time.

rust/noosphere-ucan/Cargo.toml
cdata: It may be vestigial. I would guess it's in a build.rs file, so if we don't have one of those then you can kill it.

rust/noosphere-ucan/Cargo.toml
cdata: I'm fine with either. Mildly prefer `web-crypto-rsa`.

PR #863, opened 2024-03-25
Update for soundness fix in the next Rust release

oli-obk:
We had to do a breaking change to fix a soundness hole in the Rust language. Your crate is affected. This should fix it and does not appear to affect any callers.

See https://github.com/rust-lang/rust/issues/122230 for more information


jsantell:
Thanks for the heads up! We have a fix for this from nightly clippy in #857 that we'll be using

